PR #24, opened 2022-06-07
Integrate WebsocketBackfiller into AlchemyWebsocketProvider

thebrianchen:
This PR adds backfiller and heartbeat behavior into the AlchemyWebsocketProvider class.
- Partially copied over from alchemy-web3
- Did some hacky work since the base `WebsocketProvider` emits some events by default.
- Added unit tests.


src/internal/alchemy-websocket-provider.ts
```diff
@@ -193,9 +341,363 @@ export class AlchemyWebSocketProvider
...
+      'down',
+      this.stopHeartbeatAndBackfill.bind(this)
+    );
+  }
+
+  private removeSocketListeners(): void {
+    this._websocket.removeEventListener('message', this.handleMessage);
+    this._websocket.removeEventListener('reopen', this.handleReopen);
+    this._websocket.removeEventListener('down', this.stopHeartbeatAndBackfill);
+  }
```
dphilipson, line 358: This removal code won't work: you need to pass the same callback instance to `removeEventListener` that was originally passed to `addEventListener`, but here the callback added is a different instance because `.bind` returns a new instance.

One way to avoid this is to leave out the `.bind` and make sure that all three of `handleMessage`, `handleReopen`, and `stopHeartbeatAndBackfill` are bound methods. That is, they should all be written in the style:
```ts
private handleMessage = (event: MessageEvent): void => {
  // ...
}
```
and not in the style
```ts
private handleMessage(event: MessageEvent): void {
  // ...
}
```
At the moment, it appears that `handleMessage` is bound but the other two are not.

If you find that syntax too finicky, you can also store `this.handleMessage.bind(this)`, etc. in fields so you can reference the same instance in both add and remove.

src/internal/alchemy-websocket-provider.ts
```diff
@@ -184,6 +302,36 @@ export class AlchemyWebSocketProvider
...
 
+  /** @internal */
+  async sendBatch(parts: BatchPart[]): Promise<any[]> {
+    let nextId = 0;
+    const payload: JsonRpcRequest[] = parts.map(({ method, params }) => {
+      return {
+        method,
+        params,
+        jsonrpc: '2.0',
+        id: `alc-web3:${nextId++}`
```
dphilipson, line 313: (Nit) May want to change the id from `alc-web3` to something else.

dphilipson (commented):
Wow, nice work on this! Some minor comments, but seems like it should work great! Thanks for putting in all the effort to make this work.

thebrianchen (commented):
thanks for detailed review!

dphilipson (approved):
Looks great! Thanks for putting all the time to make this work!

PR #25, opened 2022-06-08
Support array types in NFT descriptions

thebrianchen:
Some NFTs have a their `description` fields as an array. The SDK should handle this appropriately and join the array as a string.

Example contract: `0x0075ce3f8d2cecae84600d4a1f8675a72819896d`

### DIFF
--- a/src/api/nft.ts
+++ b/src/api/nft.ts
@@ -35,13 +35,13 @@ export class BaseNft {
   }
 
   /** @internal */
-  static fromResponse(ownedNft: RawBaseNft, contractAddress: string): BaseNft {
+  static fromResponse(baseNft: RawBaseNft, contractAddress: string): BaseNft {
     return new BaseNft(
       contractAddress,
       // We have to normalize the token id here since the backend sometimes
       // returns the token ID as a hex string and sometimes as an integer.
-      BigNumber.from(ownedNft.id.tokenId).toString(),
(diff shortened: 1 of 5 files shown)


test/unit/nft.test.ts
```diff
@@ -88,6 +88,22 @@ describe('Nft class', () => {
...
+      'title',
+      tokenIdString,
+      NftTokenType.ERC721,
+      { raw: '', gateway: '' },
+      [],
+      '2022-02-16T17:12:00.280Z',
+      description
+    );
+    const nft = Nft.fromResponse(rawNft, '0xCA1');
+    expect(nft.description).toEqual('very special and unique description');
```
ddxgod, line 104: noice

src/api/nft.ts
```diff
@@ -105,23 +105,30 @@ export class Nft extends BaseNft {
...
+      nft.metadata,
+      nft.error
     );
   }
 
+  /** @internal */
+  private static parseDescription(description: string | Array<string>): string {
+    return typeof description === 'string'
+      ? description
+      : description.join(' ');
```
ddxgod, line 129: random nit, stripping the description of whitespaces? not sure if that's necessary

ddxgod (approved):

src/api/nft.ts
thebrianchen: I'm not stripping the whitespace here, just adding a space in between each array element since the example NFT doesn't add any spacing.

PR #27, opened 2022-06-08
Add getOwnersForCollection endpoint + cleanup README

thebrianchen:
Adding `getOwnersForCollection()` endpoint along with some README cleanup.
- Note that developer return type is `owners` whereas the backend returns it under `ownerAddresses`. This is done to standardize with `getOwnersForNft()`, which also returns as `owners`.


src/api/nft-api.ts
```diff
@@ -264,7 +266,7 @@ export async function getNftsForCollection(
 ): Promise<CollectionNftsResponse | CollectionBaseNftsResponse> {
   const withMetadata = omitMetadataToWithMetadata(options?.omitMetadata);
   const response = await requestHttpWithBackoff<
-    GetNftsForCollectionAlchemyParams,
+    GetOwnersForCollectionAlchemyParams,
```
alex-miao, line 269: Why is this within `getNftsForCollection`? Think you want a new function `getOwnersForCollection`.

src/api/nft-api.ts
```diff
@@ -610,7 +661,7 @@ function omitMetadataToWithMetadata(
  *
  * @internal
  */
-interface GetNftsForCollectionAlchemyParams {
+interface GetOwnersForCollectionAlchemyParams {
   contractAddress: string;
   startToken?: string;
```
alex-miao, line 666: No need for `startToken` and `withMetadata` for `getOwnersForCollection`.

alex-miao (changes requested):

src/api/nft-api.ts
thebrianchen: oops, forgot to rename back. Thanks for catching!

alex-miao (commented):
Code changes look good! Getting a CircleCI build error but will stamp once that gets fixed.

alex-miao (approved):
Approving since CircleCI fix is separate PR.

PR #37, opened 2022-06-27
Add getNftContractMetadata

deric-alchemy:
Need to add tests still! But a quick code review would be great

### DIFF
--- a/src/api/nft-api.ts
+++ b/src/api/nft-api.ts
@@ -16,7 +16,7 @@ import {
 } from '../types/types';
 import { Alchemy } from './alchemy';
 import { paginateEndpoint, requestHttpWithBackoff } from '../internal/dispatch';
-import { BaseNft, Nft } from './nft';
+import { BaseNft, BaseNftContract, Nft, NftContract } from './nft';
 import {
   RawBaseNft,
   RawCollectionBaseNft,
@@ -27,6 +27,7 @@ import {
   RawGetNftsResponse,
   RawGetOwnersForCollectionResponse,
   RawNft,
+  RawNftContract,
   RawOwnedBaseNft,
   RawOwnedNft
 } from '../internal/raw-interfaces';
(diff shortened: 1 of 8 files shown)


src/api/nft.ts
```diff
@@ -1,15 +1,69 @@
...
+  RawBaseNft,
+  RawBaseNftContract,
+  RawNft,
+  RawNftContract
+} from '../internal/raw-interfaces';
 import { BigNumber } from 'ethers';
 
+export interface NftContractMetadata {
+  /** The name of the contract. */
+  name: string;
```
thebrianchen, line 12: These fields should be marked optional since docs state that the fields are only set if the contract implements them.

`tokenType` should always be defined for the end-developer, since it should default to `UNKNOWN` if not specified by the backend

src/api/nft.ts
```diff
@@ -1,15 +1,69 @@
...
+
 /**
- * Represents an NFT contract.
+ * Alchemy representation of a base NFT contract that doesn't contain metadata.
  *
  * @public
  */
-export interface NftContract {
-  /** The NFT contract address. */
+export class BaseNftContract {
```
thebrianchen, line 26: I think it'd be cleaner if these were just interfaces instead of classes, like `TokenUri`. Just for context, I originally had `Nft` and `BaseNft` as classes since I wanted to eventually add helper methods onto it. In this case, having a simple interface would be cleaner.

Also, as discussed, can we move contractMetadata fields up for better ergonomics? Here's what I'm roughly thinking
```ts
export interace BaseNftContract {
  address: string
}

export interface NftContract extends BaseNftContract {
  name?: string
  symbol?: string
  ...
}
```

src/api/nft-api.ts
```diff
@@ -694,6 +745,17 @@ interface GetNftMetadataParams {
   refreshCache?: boolean;
 }
 
+/**
+ * Interface for the `getNFTContractMetadata` endpoint.
+ *
+ * @internal
+ */
+interface GetNFTContractMetadataParams {
```
thebrianchen, line 753: This should just be `contractADdress` only right?

thebrianchen (commented):
Thanks for getting this out so fast! Left some comments/nits, feel free to push back or discuss more!

src/api/nft.ts
deric-alchemy: Can lift contractMetadata, no problem!

For the interface vs. classes, don't we need the class to implement "fromResponse" to parse from RawBaseNFTContract?

deric-alchemy: Moved to utils.ts!

deric-alchemy:
need formatting for some files

PR #39, opened 2022-06-27
Refactor NFT classes to be interfaces, create util functions

deric-alchemy:


src/api/util.ts
```diff
@@ -29,3 +32,55 @@ export function toHex(num: number): string {
 export function isHex(possibleHexString: string): boolean {
   return /^0x[0-9a-fA-F]+$/.test(possibleHexString);
 }
+
+export function getBaseNftFromRaw(
```
thebrianchen, line 36: Can you move these to `src/util/util.ts`? I'm hoping to leave all methods in `src/api` as user-facing exports. 

Feel free to rename the other file to make it less confusing!

src/api/util.ts
```diff
@@ -29,3 +32,55 @@ export function toHex(num: number): string {
...
+  rawBaseNft: RawBaseNft,
+  contractAddress: string
+): BaseNft {
+  return {
+    contract: { address: contractAddress },
+    tokenId: BigNumber.from(rawBaseNft.id.tokenId).toString(),
```
thebrianchen, line 42: nit: can you move the comments that were originally on the class over to the util as well?

thebrianchen (approved):
lgtm! thanks for taking on the refactor

src/api/util.ts
deric-alchemy: done!

PR #41, opened 2022-06-28
Added isSpamContract

deric-alchemy:


src/types/types.ts
```diff
@@ -328,6 +328,35 @@ export interface GetOwnersForCollectionResponse {
   readonly owners: string[];
 }
 
+export interface FloorPriceMarketplace {
```
thebrianchen, line 331: can you add a top-level comment for what a `FloorPriceMarketplace` is?

src/types/types.ts
```diff
@@ -435,6 +464,17 @@ export interface CollectionNftsResponse {
   pageKey?: string;
 }
 
+/**
+ * The response object for the {@link getSpamContracts} function. The object
+ * contains a list of NFT spam contracts as marked by Alchemy.
+ *
+ * @public
+ */
+export interface GetSpamContractsResponse {
```
thebrianchen, line 473: How do you feel about just returning an array with the addresses (`getSpamContracts(): Promise<string[]>`)?

This way, there's no need to call `.spamContracts` on the returned response. Only reason I can see otherwise is if the NFT API team wants to return an object to add more fields in the future.

thebrianchen (approved):
Looks good! Just one comment about whether we want to return the spam contracts directly as an array or inside an object.

Also please resolve conflicts before merging, thansk!

src/types/types.ts
deric-alchemy: Sounds good to me!

PR #42, opened 2022-06-30
Use 'websocket' library instead of 'ws' for easier integration with create-react-app

thebrianchen:
The default 'ws' library requires a lot of polyfills in create-react-app. Let's switch to the 'websocket' library since it's a lot easier to get started with. If package size becomes a future problem, we can revisit using 'ws' again.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@alch/alchemy-sdk",
-  "version": "1.0.3",
+  "version": "1.0.4",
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -86,6 +86,7 @@
     "typedoc-plugin-markdown": "^3.12.1",
     "typescript": "^4.5.5"
   },
+  "sideEffects": false,
```
jaypaik, line 89: Curious if this helped reduce bundle size.

PR #45, opened 2022-07-01
Fix imports to reduce bundle size in treeshaking

thebrianchen:
Trying a few things to reduce bundle size:
- Changed the `ethers` import to `@ethersproject` and updated `package.json` accordingly
- Marked `@ethersproject/providers/lib/base-provider.ts` as external dep in rollup to avoid it getting bundled in core `index.esm.js` output

Open questions:
- Is there some way to import `lib.esm/base-provider` in browser environments and `lib/base-provider` in node environments? `lib.esm` results in a smaller bundle size, but isn't compatible with the node distribution.

Huge shoutout to @dphilipson and @jaypaik for figuring most of this out!

### Importing only the `Network` enum
Before:
Screen Shot 2022-07-01 at 3 37 32 PM

After:
Screen Shot 2022-07-01 at 3 35 34 PM

### Importing `initializeAlchemy()`
Before:
Screen Shot 2022-07-01 at 3 37 03 PM

After:
Screen Shot 2022-07-01 at 3 36 33 PM


src/api/alchemy-provider.ts
```diff
@@ -152,8 +155,8 @@ export class AlchemyProvider
   }
 
   /**
-   * Overrides the base {@link providers.JsonRpcProvider.send} method to
-   * implement custom logic for sending requests to Alchemy.
+   * Overrides the base {@link JsonRpcProvider.send} method to implement custom
```
jaypaik, line 158: Rationale behind keeping some `@link`s like this vs. not using it like for `WebSocketProvider` below?

thebrianchen: I'd prefer to use `@link`s where applicable since you can go to the original definition directly when using an IDE, but switched it in other places where we're not importing from the top level.

jaypaik (approved):

PR #47, opened 2022-07-06
Remove 'lib/base-provider' dependency for reduced bundle size

thebrianchen:
Further reducing bundle size by copying the `Event` class over directly from ethers instead of using the external dependency on `@ethersproject/providers/lib/base-provider`.

Thanks @dphilipson for this realization!

### Before
Screen Shot 2022-07-06 at 12 02 02 PM

### After
Screen Shot 2022-07-06 at 12 00 54 PM

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -22,19 +22,8 @@ const allBuilds = {
       sourcemap: true
     }
   ],
-  external: [
-    ...Object.keys(pkg.dependencies || {}),
-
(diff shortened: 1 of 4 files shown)


jaypaik (approved):

jaypaik (commented):
[nice](https://app.graphite.dev/macros/alchemyplatform)

Also might be worth opening an issue/submitting a PR to ethers at some point.

PR #54, opened 2022-07-11
Move AlchemyProvider and AlchemyWebSocketProvider methods to top level

thebrianchen:
Moving methods up from `AlchemyProvider` and `AlchemyWebSocketProvider` to `Alchemy` instance. Some methods for `AlchemyWebSocketProvider` had to be moved up as well.

### AlchemyProvider
- `getBalance()`
- `getCode()`
- `getStorageAt()`
- `getTransactionCount()`
- `getBlock()`
- `getBlockWithTransactions()`
- `getNetwork()`
- `getGasPrice()`
- `getFeeData()`
- `ready`
- `call()`
- `estimateGas()`
- `getTransaction()`
- `getTransactionReceipt()`
- `sendTransaction()`
- `waitForTransaction()`
- `getLogs()`
- `send()`
### AlchemyWebSocketProvider
- `on()`
- `once()`
- `off()`
- `removeAllListeners()`
- `listeners()`

## List of methods that were not moved
- ENS methods: `getAvatar()`, `getResolver()`, `lookupAddress()`, `resolveName()` -- We should delay moving these until we decide if we want to do a custom ENS resolution endpoint
- WebSocket methods: `emit()` -- Internal methods
- Inspection methods: `isProvider()` -- There's no confusion here since we only use one provider
- base provider: `polling()`, `pollingInterval()` -- We don't expose a polling subscription option, which makes these redundant. We can explore adding this w/ enough developer feedback
- JsonRpcProvider: `getSigner()`, `getUncheckedSigner()`, `listAccounts()` -- Haven't looked deeply into what this entails
- all other internal properties/methods that are accessible on the object, but not documented on ethers are omitted

### DIFF
--- a/src/api/alchemy-websocket-provider.ts
+++ b/src/api/alchemy-websocket-provider.ts
@@ -115,7 +115,7 @@ export class AlchemyWebSocketProvider
   }
 
   /**
-   * Overridden implementation of ethers' that includes Alchemy based subscriptions.
+   * Overridden implementation of ethers that includes Alchemy based subscriptions.
    *
    * @param eventName Event to subscribe to
    * @param listener The listener function to call when the event is triggered.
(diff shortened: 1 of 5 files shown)


src/api/alchemy-websocket-provider.ts
```diff
@@ -675,6 +755,128 @@ export class AlchemyWebSocketProvider
         throw new Error('Invalid event type to `emitProcessFn()`');
     }
   }
+
```
thebrianchen, line 758: I am very much not a fan of all this copy/paste, but since `getEventTag()` isn't exposed, I can't really think of another way to support Alchemy's subscription APIs. @dphilipson any advice on how to clean this up?

src/api/alchemy.ts
```diff
@@ -507,9 +525,395 @@ export class Alchemy {
...
+  /**
+   * Adds a listener to be triggered for each {@link eventName} event. Also
+   * includes Alchemy's Subscription API events. See {@link AlchemyEventType} for
+   * how to use them.
+   *
+   * @param eventName The event to listen for.
+   * @param listener The listener to call when the event is triggered.
+   * @public
+   */
+  async on(
```
thebrianchen, line 828: One problem we now run into with dynamic imports is that we can't maintain the same API surface of `on(): AlchemyWebSocketProvider` anymore. I don't really see developers chaining these calls like `alchemy.on().off()` since most of these are meant to be one-off methods.

I'm okaaaay-ish with this based on the use case, but I wonder if you see any way to make this nicer @dphilipson. Thanks!

dphilipson: Isn't the original behavior to return `this`? What's stopping us from doing that? E.g.
```ts
  on(
    eventName: AlchemyEventType,
    listener: Listener
  ): this {
    (async () => {
      const provider = await this.getWebsocketProvider();
      provider.on(eventName, listener);
    })();
    return this;
  }
```

dphilipson (approved):
Looks good! Love the simple interface.

src/api/alchemy.ts
```diff
@@ -507,9 +525,395 @@ export class Alchemy {
...
+   *   Defaults to 'latest' if unspecified.
+   * @public
+   */
+  async getBalance(
+    addressOrName: string | Promise<string>,
+    blockTag?: BlockTag | Promise<BlockTag>
+  ): Promise<BigNumber> {
+    const provider = await this.getProvider();
+    return provider.getBalance(addressOrName, blockTag);
+  }
```
dphilipson, line 541: If you don't like writing these repetitive methods that pass through to the provider, there's an option to do tricks with a dynamic proxy to save writing. But this way is good too because it's straightforward.

src/api/alchemy-websocket-provider.ts
dphilipson: I don't know a good alternative unfortunately. Seems like we're stuck: we need to modify part of the function that isn't designed for extension, so nothing we can do other than replace the function entirely.

src/api/alchemy.ts
```diff
@@ -525,9 +929,17 @@ export class Alchemy {
...
+   * `AlchemyWebSocketProvider` class and has been expanded to support Alchemy's
+   * Subscription APIs, automatic backfilling, and other performance improvements.
+   *
+   * Most common methods on the provider are available as top-level methods on
+   * the {@link Alchemy} instance, but the provider is exposed here to access
+   * other less-common methods.
+   *
    * @public
    */
   async getWebsocketProvider(): Promise<AlchemyWebSocketProvider> {
```
dphilipson, line 945: Yo, I just noticed: isn't this subject to race conditions? If someone were to call this multiple times before the import completes, we'd end up creating multiple websocket providers, which is an issue because we'll end up with multiple connections.

Seems like we want something like
```ts
private wssProviderPromise: Promise<AlchemyWebSocketProvider> | undefined;

getWebsocketProvider(): Promise<AlchemyWebSocketProvider> {
  if (!this.wssProviderPromise) {
    this.wssProviderPromise = (async () => {
      const { AlchemyWebSocketProvider } = await import(
        './alchemy-websocket-provider'
      );
      return new AlchemyWebSocketProvider(
        this.network,
        this.apiKey
      );
    })();
  }
  return this.wssProviderPromise;
}

}
```

src/api/alchemy.ts
thebrianchen: > What's stopping us from doing that? E.g.
Doh... only my ignorance. In my mind I thought needed to return the underlying provider rather than the top level. thanks!

src/api/alchemy.ts
thebrianchen: wow, that's a good catch, thanks! I guess the average person won't be calling `getProvider()` without an `await` in front of it, but that's still a bad bug.

Added a test that verifies we don't race.

dphilipson: Yea, you're right most people likely won't hit it. It's a bit more likely now that we have top-level methods, so someone could be like
```ts
const alchemy = new Alchemy(stuff);
alchemy.on("blocks", () => { /* ... */ });
alchemy.on("logs", () => { /* ... */ });
```

dphilipson (approved):

src/api/alchemy.ts
```diff
@@ -918,12 +921,10 @@ export class Alchemy {
    */
   async getProvider(): Promise<AlchemyProvider> {
```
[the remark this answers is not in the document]
thebrianchen: thanks for catching

PR #55, opened 2022-07-12
Make changes to new subscription types

thebrianchen:
This PR updates the subscription types to match new [changes](https://docs.alchemy.com/alchemy/documentation/changelog#06-22-2022) to the Subscription API
- `alchemy_filteredNewFullPendingTransactions` and `alchemy_newFullPendingTransactions` have been replaced with `alchemy_pendingTransactions`
- simplified types and added logic to serialize and deserialize the custom Alchemy subscription event.
- renamed `Alchemy.getWebsocketProvider()` to `Alchemy.getWebSocketProvider()`


src/util/util.ts
```diff
@@ -98,3 +98,17 @@ function parseNftTokenUriArray(arr: TokenUri[] | undefined): TokenUri[] {
   }
   return arr.filter(uri => parseNftTokenUri(uri) !== undefined);
 }
+
+export function stripUndefined(obj: { [key: string]: any }): {
```
deric-alchemy, line 102: Can we add a comment on why this is necessary (either here or in the usage)? The low-level code is a little hard to read without context.

README.md
```diff
@@ -97,39 +97,35 @@ runAlchemy();
...
 ```ts
 import { Alchemy } from '@alch/alchemy-sdk';
 
 const alchemy = new Alchemy();
 
-const websocketProvider = alchemy.getWebsocketProvider();
-
 // Listen to all new pending transactions.
-websocketProvider.on(
+alchemy.on(
```
deric-alchemy, line 111: This syntax is honestly a lot more confusing than web3.js (with "subscribe" being quite obvious). I'll have to update our docs well to make sure it's obvious what's happening

thebrianchen: yeah, ethers has a different model of doing subscriptions, and i tried to just add the additional overload here. lmk if you can think of another way to make it clearer!

deric-alchemy: Yeah, we need to mirror Ethers if we want it to be a drop-in replacement, so it'll have to rely on documentation

src/api/alchemy-websocket-provider.ts
```diff
@@ -1008,13 +1003,57 @@ function addToPastEventsBuffer<T>(
...
+  event: AlchemyEventType
+): event is AlchemyPendingTransactionsEventFilter {
   return typeof event === 'object' && 'method' in event;
 }
 
-function getAlchemyEventTag(event: AlchemyEventType): string {
+/**
+ * Creates a string representation of an `alchemy_pendingTransaction`
+ * subscription filter that is compatible with the ethers implementation of event tag.
+ *
```
deric-alchemy, line 1015: Can we link to documentation of Ethers implementation? Or did you just have to reverse engineer this

src/api/alchemy.ts
```diff
@@ -943,7 +943,7 @@ export class Alchemy {
    *
    * @public
    */
-  getWebsocketProvider(): Promise<AlchemyWebSocketProvider> {
+  getWebSocketProvider(): Promise<AlchemyWebSocketProvider> {
```
deric-alchemy, line 946: Nice catch!

src/internal/internal-types.ts
```diff
@@ -140,15 +142,39 @@ class Event {
  * @internal
  */
 export class EthersEvent extends Event {
-  get address(): string | null {
+  get fromAddress(): string | string[] | undefined {
```
deric-alchemy, line 145: Add a comment explaining what's happening in this class as well please!

README.md
```diff
@@ -97,39 +97,35 @@ runAlchemy();
...
+    method: 'alchemy_pendingTransactions'
   },
   res => console.log(res)
 );
 
 // Listen to all transactions on the USDC contract.
-websocketProvider.on(
+alchemy.once(
   {
     method: 'alchemy_filteredNewFullPendingTransactions',
```
dphilipson, line 121: Like the above, let's also listen to `alchemy_pendingTransactions` instead of this long thing (they are aliases).

src/internal/internal-types.ts
deric-alchemy: Specifically what fields and format (toAddress, fromAddress, hashes) are being supported by this code

README.md
```diff
@@ -97,39 +97,35 @@ runAlchemy();
...
   {
-    method: 'alchemy_newFullPendingTransactions'
+    method: 'alchemy_pendingTransactions'
   },
   res => console.log(res)
 );
 
 // Listen to all transactions on the USDC contract.
-websocketProvider.on(
+alchemy.once(
```
dphilipson, line 119: Why `once`? It feels like an unusual thing to do, and doesn't match the comment.

deric-alchemy (approved):

src/api/alchemy-websocket-provider.ts
```diff
@@ -687,22 +691,16 @@ export class AlchemyWebSocketProvider
...
-          ['alchemy_newFullPendingTransactions'],
-          this.emitProcessFn(event),
-          event
-        );
-      }
+      const { fromAddress, toAddress, hashesOnly } = event;
+      void this._subscribe(
+        event.tag,
+        [
+          'alchemy_pendingTransactions',
```
dphilipson, line 698: Am I reading this correctly that we use the type "alchemy" internally to represent Alchemy pending transaction events? That feels pretty unintuitive, and also confusing if we add other Alchemy-specific streams in the future. The event type should reference transactions somehow.

src/api/alchemy-websocket-provider.ts
```diff
@@ -1008,13 +1003,57 @@ function addToPastEventsBuffer<T>(
...
+  } else {
+    return field;
+  }
+}
+
+function serializeBooleanField(field: boolean | undefined): string | undefined {
+  if (field === undefined) {
+    return '*';
+  } else {
+    return field ? 'true' : 'false';
```
dphilipson, line 1057: This is fine, but can also be written `return field.toString()` or `return field + ""`.

src/api/alchemy-websocket-provider.ts
```diff
@@ -687,22 +691,16 @@ export class AlchemyWebSocketProvider
...
-          this.emitProcessFn(event),
-          event
-        );
-      }
+      const { fromAddress, toAddress, hashesOnly } = event;
+      void this._subscribe(
+        event.tag,
+        [
+          'alchemy_pendingTransactions',
+          stripUndefined({ fromAddress, toAddress, hashesOnly })
```
dphilipson, line 699: Why's the `stripUndefined` needed? Seems like this would only matter if you plan to iterate over the keys, which I don't see happening anywhere.

dphilipson (commented):
Looks great! Some minor comments inline.

src/api/alchemy-websocket-provider.ts
thebrianchen: Good idea! changed from `alchemy` to `alchemy-pending-transactions`. Need to use lower case since ethers turns everything lowercase.

README.md
thebrianchen: oops, forgot to change it.

README.md
thebrianchen: I wanted to highlight that there is also a one-time listener that fires. Also down to keep `on()` if that makes more sense.

src/util/util.ts
thebrianchen: done.

README.md
dphilipson: That seems reasonable, but should change the comment then to emphasize it.

dphilipson: Oh nm, I see you changed this.

dphilipson: With the comment changed it looks good. 👍

src/api/alchemy-websocket-provider.ts
thebrianchen: thanks for prompting me to dig deeper! I originally had tests fail since undefined fields were present, but I realize that the test was stubbing out `send()`, rather than the underlying websocket server. Removed the method entirely.

PR #70, opened 2022-07-21
Split SDK into namespaces

thebrianchen:


src/api/nft-namespace.ts
```diff
@@ -0,0 +1,646 @@
...
+
+    return {
+      contractAddress: response.contractAddress,
+      refreshState: parseReingestionState(response.reingestionState),
+      progress: response.progress
+    };
+  }
+}
+
+async function refresh(
```
deric-alchemy, line 479: Should we make this private? Assuming it's an internal method

thebrianchen: The function isn't exported and isn't a member function on the namespace class, so we're fine.

src/api/nft-namespace.ts
```diff
@@ -0,0 +1,646 @@
...
+    return getNftFromRaw(response, contractAddress);
+  }
+
+  /**
+   * Get the NFT collection metadata associated with the provided parameters.
+   *
+   * @param contractAddress - The contract address of the NFT.
+   * @public
+   */
+  async getNftContractMetadata(contractAddress: string): Promise<NftContract> {
```
deric-alchemy, line 80: This should be getContractMetadata post-namespace!

thebrianchen: done

PR #81, opened 2022-07-27
Pass method names in header for NFT API

thebrianchen:
This PR adds the original SDK method name into the header.
- To avoid exposing the method name overloads to the end developer, we need to move the methods into an internal wrapper.
- I ended up moving all of them at once even though some are direct mappings, since we'll eventually want to add request IDs into the headers as well for all methods.


deric-alchemy (approved):

deric-alchemy:
What does the header look like with the original method name?

What will it look like with request IDs?

thebrianchen:
> What does the header look like with the original method name?
> 
> What will it look like with request IDs?

Example call for `checkOwnership()` would be: 
```
headers: {
  'Alchemy-Ethers-Sdk-Version': '2.0.0';
  'Alchemy-Ethers-Sdk-Method': 'checkOwnership'
},
params: {
  method: 'getNFTs'
  ...
}
```

Request ids are still further down on the roadmap, but if we added a group by request Id, it would likely be an additional header

dphilipson (approved):

PR #83, opened 2022-07-29
Add send() override in AlchemyProvider()

thebrianchen:
This PR adds the method header for Enhanced APIs and `send()`. This is not the most optimal since this doesn't allow tracking some core namespace calls, but opens the door for a future refactor.

Changes:
- copy over `JsonRpcProvider.send()` as `_send()` in order to add add header into the `ConnectionInfo` (this avoids needing to copy everything in the `fetchJson()` chain of calls.
- fix a bug where providing the URL doesn't pass in the network

todo for another PR:
- we can get more accurate method headers by overriding `perform()` to pass the method header, but we can postpone that until we need to capture method headers for one of those methods.

Tested manually and verified that headers are landing in imply.

### DIFF
--- a/src/api/alchemy-provider.ts
+++ b/src/api/alchemy-provider.ts
@@ -6,7 +6,8 @@ import {
   Network as NetworkFromEthers,
   Networkish
 } from '@ethersproject/networks';
-import { ConnectionInfo } from '@ethersproject/web';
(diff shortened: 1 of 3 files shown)


deric-alchemy (approved):

deric-alchemy:
Looks good!

PR #94, opened 2022-08-12
Add support for ARB_GOERLI 

thebrianchen:
This PR adds support for the Arbitrum-Goerli Network.

Since the underlying ethers.js provider doesn't have support for this network yet, this PR adds the override to the `static getNetwork()` function that ethers uses to verify the network. Any subsequent network overrides can just be added into the `getNetwork()` calls on both the regular and websocket providers.

Also added some unit and integration tests to verify the e2e behavior of each network for all the providers.

### DIFF
--- a/src/api/alchemy-websocket-provider.ts
+++ b/src/api/alchemy-websocket-provider.ts
@@ -4,7 +4,8 @@ import { AlchemyProvider } from './alchemy-provider';
 import { Listener } from '@ethersproject/abstract-provider';
 import {
   AlchemyEventType,
-  AlchemyPendingTransactionsEventFilter
+  AlchemyPendingTransactionsEventFilter,
+  Network
 } from '../types/types';
 import {
   BatchPart,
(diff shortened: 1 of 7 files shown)


src/api/alchemy-websocket-provider.ts
```diff
@@ -120,6 +130,24 @@ export class AlchemyWebSocketProvider
...
+  /**
+   * Overrides the `BaseProvider.getNetwork` method as implemented by ethers.js.
+   *
+   * This override allows the SDK to set the provider's network to values not
+   * yet supported by ethers.js.
+   *
+   * @internal
+   * @override
+   */
+  static getNetwork(network: Networkish): NetworkFromEthers {
```
deric-alchemy, line 142: Possibly make this a private util class to avoid duplicate code?

deric-alchemy (approved):

deric-alchemy:
Looks good!

src/api/alchemy-websocket-provider.ts
thebrianchen: This needs to be public so that the super constructor code ends up finding this override ([context](https://github.com/ethers-io/ethers.js/blob/master/packages/providers/src.ts/json-rpc-provider.ts#L432): this is where it traverses the object to find the first `static` call to this method)

PR #98, opened 2022-08-17
Added Contract to alchemy-sdk from ethers

deric-alchemy:
Ready to go!


src/api/alchemy-contract.ts
```diff
@@ -0,0 +1,3 @@
+import { Contract as EthersContract } from '@ethersproject/contracts';
+
+export class Contract extends EthersContract {}
```
dphilipson, line 3: The subclass here is because we plan to later add or tweak stuff in it?

thebrianchen: we're going to eventually add overloads to pass in the `Alchemy` instance into the constructor instead of the needing to pass in the provider

thebrianchen (approved):

PR #99, opened 2022-08-17
Support `sendGasOptimizedTransaction()`

thebrianchen:
This PR adds support for multi-payload transactions and automatic gas-optimized transactions. To support these features, we've also added:
- support for `eth_maxPriorityFeePerGas`
- moved `estimateGas()` into the `transact` namespace


src/api/transact-namespace.ts
```diff
@@ -134,4 +300,49 @@ export class TransactNamespace {
...
+      gasLimit,
+      maxFeePerGas: Math.round(
+        baseFee * feeMultiplier + priorityFee * feeMultiplier
+      ),
+      maxPriorityFeePerGas: Math.round(feeMultiplier * priorityFee)
+    };
+    gasTransactions.push(txWithGas);
+  }
+
+  return gasTransactions;
```
dphilipson, line 347: You could do
```ts
return GAS_OPTIMIZED_TX_FEE_MULTIPLES.map((feeMultiplier) => {
  // ...
});
```
for this if you wanted.

dphilipson (approved):

src/api/transact-namespace.ts
thebrianchen: nice! thanks for spotting that one

PR #105, opened 2022-08-25
Update Wallet to use the Alchemy object

deric-alchemy:


deric-alchemy:
Note - this is a breaking change :( But hopefully not a heavy-duty one, since we just added the Wallet.

deric-alchemy:
Added constructor overload!

src/api/alchemy-wallet.ts
```diff
@@ -1,3 +1,105 @@
...
+  async resolveName(name: string): Promise<string> {
+    return this.getWallet().then(wallet => wallet.resolveName(name));
+  }
+
+  private async getWallet() {
+    if (!this.hasAlchemyObject) {
+      return this.connect(this.provider);
+    }
+    if (!this.alchemy) {
+      throw new Error('No Alchemy object passed into the wallet.');
```
dphilipson, line 95: This case seems impossible, since `hasAlchemyObject` and `alchemy` are always set together.

Actually, do we need a separate `hasAlchemyObject` field? Seems like we can just check if `this.alchemy` is present.

src/api/alchemy-wallet.ts
```diff
@@ -1,3 +1,105 @@
...
+    }
+    if (!this.alchemyProvider) {
+      this.alchemyProvider = await this.alchemy.config.getProvider();
+    }
+    if (!this.alchemyProvider) {
```
dphilipson, line 100: This case also seems impossible: the type signature of `getProvider` indicates that it can never resolve to a falsy value.

test/unit/wallet.test.ts
```diff
@@ -41,4 +51,58 @@ describe('Alchemy-Ethers Wallet', () => {
...
+  it('Matches functionality of Ethers Signer', async () => {
+    const alchProvider = await alchemy.config.getProvider();
+
+    const alchWallet = new Wallet(privateKey, alchemy);
+    const ethersWallet = new EthersWallet(privateKey, alchProvider);
+    const blockTag = 15000000;
+
+    expect(await alchWallet.getBalance(blockTag)).toEqual(
+      await ethersWallet.getBalance(blockTag)
+    );
```
dphilipson, line 64: This is fine, but fyi you can also write these as:
```ts
await expect(alchWallet.getBalance(blockTag)).resolves.toEqual(
  await ethersWallet.getBalance(blocktag)
);
```
It's more verbose, but it makes a better error message in the case where the promise fails.

src/api/alchemy-wallet.ts
```diff
@@ -1,3 +1,105 @@
...
+  private async getWallet() {
+    if (!this.hasAlchemyObject) {
+      return this.connect(this.provider);
+    }
+    if (!this.alchemy) {
+      throw new Error('No Alchemy object passed into the wallet.');
+    }
+    if (!this.alchemyProvider) {
+      this.alchemyProvider = await this.alchemy.config.getProvider();
+    }
```
dphilipson, line 99: This could be improved: as written, if the user makes multiple calls at once, then we could end up calling `getProvider()` multiple times, which can have a real cost associated with if if the provider is a WebSocket provider because it will open multiple WebSocket connections that will never be closed.

Instead, consider structuring this class as follows, keeping a single promise around so that multiple calls before the provider is created end up getting the same provider:
```ts
export class Wallet extends EthersWallet {
  private alchemyProviderPromise?: Promise<Provider>;

  // ...

  private async getWallet(): Promise<Provider> {
    // ...
    if (!this.alchemyProviderPromise) {
      this.alchemyProviderPromise = this.alchemy.config.getProvider();
    }
    return this.connect(await this.alchemyProviderPromise);
  }
}
```

dphilipson (commented):
This looks great! I love the ergonomics of it!

A few comments inline- the most relevant one is the one about simultaneous calls to `getWallet()`.

src/api/alchemy-wallet.ts
deric-alchemy: makes sense - cleaned up!

src/api/alchemy-wallet.ts
```diff
@@ -1,3 +1,105 @@
...
+  TransactionResponse
+} from '@ethersproject/abstract-provider';
 import { Wallet as EthersWallet } from '@ethersproject/wallet';
+import { BytesLike } from '@ethersproject/bytes';
+import { ExternallyOwnedAccount } from '@ethersproject/abstract-signer';
+import { SigningKey } from '@ethersproject/signing-key';
+import { Deferrable } from '@ethersproject/properties';
+import { Alchemy } from './alchemy';
+import { BigNumber } from '@ethersproject/bignumber';
+export class Wallet extends EthersWallet {
```
thebrianchen, line 15: can you add doc comments on the top level describing what the differences vs. regular `Wallet` exported by ethers is?

thebrianchen (commented):
a few drive by comments on docs for the end user. 

looks super clean now! thanks for taking this on

src/api/alchemy-wallet.ts
deric-alchemy: It's been removed!

src/api/alchemy-wallet.ts
deric-alchemy: Done!

src/api/alchemy-wallet.ts
```diff
@@ -1,3 +1,105 @@
...
+  readonly hasAlchemyObject: boolean = false;
```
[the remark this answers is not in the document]
deric-alchemy: done!

src/api/alchemy-wallet.ts
```diff
@@ -1,3 +1,105 @@
...
+      this.hasAlchemyObject = true;
+      this.alchemy = alchemyOrProvider;
+    }
+  }
+
+  //////////////////////////////////////////////////////////////////
+  // Set of overrides from Signer to handle async provider retrieval.
+  //////////////////////////////////////////////////////////////////
+
+  async getBalance(blockTag?: BlockTag): Promise<BigNumber> {
```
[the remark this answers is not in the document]
deric-alchemy: None are in the original Ethers class!

test/unit/wallet.test.ts
```diff
@@ -41,4 +51,58 @@ describe('Alchemy-Ethers Wallet', () => {
     const rawTx = await wallet.signTransaction(transaction);
     expect(rawTx).toEqual(expectedRawTx);
   });
+
+  it('Matches functionality of Ethers Signer', async () => {
```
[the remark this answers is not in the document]
deric-alchemy: Alchemy and Provider classes are optional - if none are provided, it will just sign the transaction but not send it!

src/api/alchemy-wallet.ts
```diff
@@ -1,3 +1,105 @@
...
+    if (alchemyOrProvider && !Provider.isProvider(alchemyOrProvider)) {
+      this.alchemyProviderPromise = alchemyOrProvider.config.getProvider();
+    }
+  }
+
+  //////////////////////////////////////////////////////////////////
+  // Set of overrides from Signer to handle async provider retrieval.
+  //////////////////////////////////////////////////////////////////
+
+  getBalance(blockTag?: BlockTag): Promise<BigNumber> {
```
thebrianchen, line 53: can you add the code comments from https://docs.ethers.io/v5/api/signer/#Signer-getChainId and add the `@param` annotations so that the docs generator is complete?

thebrianchen (approved):
lgtm, please just finish copying comments over for IDE users!

dphilipson (approved):

PR #110, opened 2022-09-07
Fix totalSupply typing bug

thebrianchen:
This fixes the payload type to match what is actually returned by the backend. I want to double check here, but I don't think this requires a major version bump since we're fixing a bug.


dphilipson (approved):

dphilipson:
Pragmatically, I think it's not worth counting typing changes as breaking changes. This is because literally any change that improves typing is "breaking" in the sense that now things that previously passed type checking no longer do: this is the whole point of improving typing.

You'll notice that TypeScript itself follows this convention: many versions of typescript contains "breaking" changes in the sense that things that previously passed typechecks no longer do, but they only rarely update the major version (which is currently only v4).

PR #116, opened 2022-09-13
Add support for 'erc20' in getTokenBalances()

thebrianchen:
This PR adds support for the 'erc20' as an option in `getTokenBalances()` along with corresponding overloads to reflect the optional `pageKey` response and param exclusive to the `erc20` option. This also changes the documentation + default overloads to reflect the endpoint returning `erc20` tokens by default when no 2nd parameter is specified.

Holding off on merging until @dzou confirms that erc20 tokens are returned by default.

Would love feedback on the overloads and if there's any way to simplify some of the logic.


dphilipson (approved):

dzou (approved):
Awesome!

PR #119, opened 2022-09-20
update getOwnersForContract

ayang-rkjs:
## add docs

PR #120, opened 2022-09-20
add isHolderOfCollection

ayang-rkjs:


ayang-rkjs:
Current dependencies on/for this PR:
* main
  * **PR #119** <a href="https://app.graphite.dev/github/pr/alchemyplatform/alchemy-sdk-js/119" target="_blank">Graphite</a> 
    * **PR #120** <a href="https://app.graphite.dev/github/pr/alchemyplatform/alchemy-sdk-js/120" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/alchemyplatform/alchemy-sdk-js/120?utm_source=stack-comment).

src/api/nft-namespace.ts
```diff
@@ -338,4 +339,20 @@ export class NftNamespace {
...
+  /**
+   * Returns 'true' if the input wallet address holds a token from the input NFT
+   * contract, and otherwise returns 'false'. This API is useful for simple
+   * token-gating use cases.
+   *
+   * @param wallet - The wallet address to check.
+   * @param contractAddress - The contract address of the NFT collection.
+   * @beta
+   */
+  isHolderOfCollection(
```
deric-alchemy, line 352: We typically use "Contract" instead of "Collection" in our SDK NFT API. @thebrianchen , can we get some thoughts on whether we should call it "Collection" or "Contract"?

Shouldn't we have called our public NFT API "isHolderOfContract" not "isHolderOfCollection"? Feels like we should have a better process for vetting these names if this slipped through the cracks.

ayang-rkjs: Talked to Brian about this and realize the SDK already has a method like this, so can close this PR.

So going forward, is the ask that all NFT APIs use the term 'contract' instead of 'collection' to refer to NFT contracts?

We can easily support both in a non-breaking way for backwards compatibility.

thebrianchen:
Closing since we already have `checkNftOwnership()`

Issue #118, opened 2022-09-20
Throws when NFT description is an object


alexvobo:
https://github.com/alchemyplatform/alchemy-sdk-js/blob/519f0c2e9900a064f0cb23f70f75a1c07a6bdd0f/src/util/util.ts#L95

This throws a type error when the description is an object because `.join(' ')` can only be called on an array.

Specifically, when I'm trying to call `.getNftsForOwner(address)` and there's an NFT where `description`:`{...}`

One NFT will prevent the rest from fetching.

thebrianchen:
@alexvobo Thanks for flagging this. The description field should always be a string. Do you have the owner or contract address for me to repro? Thanks!

thebrianchen:
@alexvobo Thanks for the issue and PR! I made a quick PR in #126 since your PR will also incorrectly stringify the description if the field is `string[]` type. This should be fixed in the next release!

PR #121, opened 2022-09-20
Handle bad input in parseNftDescription

alexvobo:
Refer to https://github.com/alchemyplatform/alchemy-sdk-js/issues/118

If an NFT with a malformed description property (`object`, when a `string` or `array` is expected) gets sent to an address, it will prevent fetching of all NFTs for that address.

We can fix it by returning an empty string, stringifying the object, or setting description to an error message.


thebrianchen:
Closing in favor of #126. Thanks for the debugging and submission!

PR #123, opened 2022-09-21
Ay/spam info

ayang-rkjs:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "alchemy-sdk",
-  "version": "2.0.4",
+  "version": "2.0.5-rc1",
```
thebrianchen, line 3: Can you undo this change?

src/types/types.ts
```diff
@@ -412,6 +412,14 @@ export interface Media {
   size?: number;
 }
 
+/** Detailed information on whether and why an NFT contract was classified as spam. */
+export interface SpamInfo {
+  isSpam: boolean;
+
+  /** A list of reasons why an NFT contract was marked as spam. */
+  classifications: string[];
```
thebrianchen, line 420: Can we turn this into a string enum with all the possible reasons? ex: https://github.com/alchemyplatform/alchemy-sdk-js/blob/347f87ac63f6fdc4d72683bf458511fe208c14fe/src/types/types.ts#L703

src/util/util.ts
```diff
@@ -65,7 +65,8 @@ export function getNftFromRaw(rawNft: RawNft, contractAddress: string): Nft {
     metadataError: rawNft.error,
     rawMetadata: rawNft.metadata,
     tokenUri: parseNftTokenUri(rawNft.tokenUri),
-    media: parseNftTokenUriArray(rawNft.media)
+    media: parseNftTokenUriArray(rawNft.media),
+    spamInfo: rawNft.spamInfo
```
thebrianchen, line 69: `isSpam` comes from the backend as a string, so we'll need to parse it to a boolean here.

src/api/nft.ts
```diff
@@ -71,4 +77,7 @@ export interface Nft extends BaseNft {
 
   /** URIs for accessing the NFT's media assets. */
   media: Media[];
+
```
thebrianchen, line 80: Can you add the `spamInfo` field to `OwnedNft` (https://github.com/alchemyplatform/alchemy-sdk-js/blob/ba8ce861b3da5a3e1dc7635b21f5da9d4137d66a/src/types/types.ts#L569) instead? That way the field is only present in `getNftsForOwner()` and not `getNftsForContract()`.

thebrianchen (changes requested):
Huge thanks for taking this on! Looks good, just a few small nits. Can we also add some [unit tests](https://github.com/alchemyplatform/alchemy-sdk-js/blob/519f0c2e9900a064f0cb23f70f75a1c07a6bdd0f/test/unit/nft-api.test.ts#L214) to make sure that raw response ends up being parsed into the correct types + an [integration test](https://github.com/alchemyplatform/alchemy-sdk-js/blob/519f0c2e9900a064f0cb23f70f75a1c07a6bdd0f/test/integration/integration.test.ts#L100) to verify the e2e behavior? tyty

package.json
ayang-rkjs: Hm I already published this to npm, should we keep it like this to reflect that?

src/api/nft.ts
ayang-rkjs: spamInfo is included in the response of getNftsForContract as well (it's a bit repetitive admittedly).

We also can't always extract this field from the raw NFT the way we do with `balance` because spamInfo is only present when we've fetched metadata (i.e. it's not present on any of the Base* types)

thebrianchen: I don't see it here: https://eth-mainnet.g.alchemy.com/nft/v2/demo/getNFTsForCollection?contractAddress=0x01234567bac6ff94d7e4f0ee23119cf848f93245&withMetadata=true

Are we planning on releasing this soon, or is it only on certain contracts?

ayang-rkjs: ah yup yup it's only present for spam contracts

src/util/util.ts
ayang-rkjs: changed the raw type and parsing

src/types/types.ts
ayang-rkjs: done!

thebrianchen (approved):
LGTM! Thanks for adding the stuff. Pls merge when you're ready :) 
Added an integration sanity test and renamed `Classification` to `NftSpamClassification` since the type is a shared top-level type that developers might end up calling directly.

PR #124, opened 2022-09-27
Add withTokenBalances and pagination to getOwnersForContract()

### LINKED ISSUE #115: getOwnersForContract Function doesn't support pagination
As per docs This API supports pagination https://docs.alchemy.com/reference/getownersforcollection,
But SDK function getOwnersForContract doesn't support pagination.

thebrianchen:
Fixes #115.

This PR adds support for `withTokenBalances` and the resulting `pageKey`.

### DIFF
--- a/src/internal/nft-api.ts
+++ b/src/internal/nft-api.ts
@@ -5,7 +5,10 @@ import {
   GetFloorPriceResponse,
   GetNftsForContractOptions,
   GetNftsForOwnerOptions,
+  GetOwnersForContractOptions,
   GetOwnersForContractResponse,
+  GetOwnersForContractWithTokenBalancesOptions,
+  GetOwnersForContractWithTokenBalancesResponse,
   GetOwnersForNftResponse,
   NftContractBaseNftsResponse,
   NftContractNftsResponse,
@@ -32,6 +35,7 @@ import {
   RawNftContract,
   RawOwnedBaseNft,
   RawOwnedNft,
+  RawOwnerAddress,
   RawReingestContractResponse
 } from './raw-interfaces';
 import { AlchemyApiType } from '../util/const';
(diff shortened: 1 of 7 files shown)


src/internal/nft-api.ts
```diff
@@ -199,17 +203,45 @@ export async function* getNftsForContractIterator(
...
+  // TODO: remove `any` cast after backend is updated
+  const response: any = await requestHttpWithBackoff<
     GetOwnersForNftContractAlchemyParams,
     RawGetOwnersForContractResponse
   >(config, AlchemyApiType.NFT, 'getOwnersForCollection', srcMethod, {
+    ...options,
     contractAddress
   });
 
+  // HACK: cast `balance` field to string until NFT API makes the change
```
deric-alchemy, line 222: Are we saving "TODO changes contingent on backend updates" anywhere in a list such that we can reference them (say, if you're out or end up on another project)?

deric-alchemy:
Why did we add withTokenBalances in the first place - customer requests? Seems like an odd feature in my head

src/internal/nft-api.ts
thebrianchen: Ah this one isn't big enough to warrant an asana task, but should be wrapped up by @ayang-rkjs soon!

thebrianchen:
> Why did we add withTokenBalances in the first place - customer requests? Seems like an odd feature in my head

The old endpoint only returned a giant list of wallet addresses, which made it hard to figure out who owned which token, or how many of tokens a wallet owned. Finding that information for multiple owners is a pain, so this option returns all the info at once for the end developer to prase.

thebrianchen:
Quick update, after discussing on slack, i'm going to revert the balance to string conversion and instead leave it as number in order to match `getNftsForOwner()`.

dphilipson (approved):

PR #129, opened 2022-09-28
Add verifyNftOwnership()

thebrianchen:
The `checkNftOwnership()` method is non-intuitive since it returns `true` if the provided owner owns any of the provided NFT addresses. This PR adds support for `verifyNftOwnership()`, which returns a single boolean value if only a single contract address is provided, or a mapping of contractAddress to the ownership result boolean if multiple contract addresses are provided. 

`checkNftOwnership()` is now marked as deprecated.


deric-alchemy (approved):
Looks good!

deric-alchemy:
Did we ever support checkNftOwnership in SDK? Are we backwards compatible here?

thebrianchen:
> Did we ever support checkNftOwnership in SDK? Are we backwards compatible here?

Yeah, `checkNftOwnership()` was just a wrapped call around `getNFTs`. The old method still works, and will just be removed with the next breaking change window.

dphilipson (approved):

PR #147, opened 2022-10-14
Add Notify Namespace

thebrianchen:
This PR adds support for CRUD operations on Notify API webhooks in a new `NotifyNamespace`. It combines several separate endpoints into method overloads to simplify the API surface and adds typings to make the request parameters and response types clearer.

This is a first draft, so please leave your thoughts on the API surface and any variable names that should be renamed. Keep in mind that all types are exported at the top-level, so we'll want to avoid naming collisions or confusion.

Still todo:
- Add unit tests.
- Move integration test keys into env vars.


src/api/notify-namespace.ts
```diff
@@ -0,0 +1,538 @@
...
+    options?: GetAddressesOptions
+  ): Promise<NftFiltersResponse>;
+  async getNftFilters(
+    webhookOrId: NftActivityWebhook | string,
+    options?: GetAddressesOptions
+  ): Promise<NftFiltersResponse> {
+    this.verifyConfig();
+    const webhookId =
+      typeof webhookOrId === 'string' ? webhookOrId : webhookOrId.id;
+    const response = await requestHttpWithBackoff<{}, RawNftFiltersResponse>(
```
dphilipson, line 150: These calls are frequent enough in this file that you might consider making a helper function that calls `requestHttpWithBackoff` but sets the options which are common in this class without having to pass them in repeatedly; that is `this.config`, `AlchemyApiType.WEBHOOK`, and the `X-Alchemy-Token` header. Not that important if something makes it annoying to do.

src/api/notify-namespace.ts
```diff
@@ -0,0 +1,538 @@
...
+    params: NftWebhookParams | AddressWebhookParams | TransactionWebhookParams
+  ): Promise<
+    | MinedTransactionWebhook
+    | DroppedTransactionWebhook
+    | NftActivityWebhook
+    | AddressActivityWebhook
+  > {
+    let appId;
+    if (
+      type == WebhookType.MINED_TRANSACTION ||
```
dphilipson, line 342: (Very nit) `===` instead of `==`.

src/api/notify-namespace.ts
```diff
@@ -0,0 +1,538 @@
...
+        throw new Error('Transaction Webhooks require an app id.');
+      }
+      appId = params.appId;
+    }
+
+    let network = NETWORK_TO_WEBHOOK_NETWORK.get(this.config.network);
+    let filters;
+    let addresses;
+    if (type === WebhookType.NFT_ACTIVITY) {
+      if (!('filters' in params) || params.filters.length === 0) {
```
dphilipson, line 355: If you want, you can write this condition as
```ts
if (!params.filters?.length)
```
but very reasonable if you think that's too tricky.

Similarly, the condition in the next branch could be written
```ts
if (!params?.addresses?.length)
```

dphilipson (approved):
Nice work, this looks really clean.

deric-alchemy:
Spent some time reviewing the API - looks good to me. Nice work simplifying the endpoints!

src/api/notify-namespace.ts
```diff
@@ -0,0 +1,538 @@
...
+      );
+    }
+  }
+}
+
+/**
+ * Mapping of webhook network representations to the SDK's network representation.
+ *
+ * @internal
+ */
```
deric-alchemy, line 464: Maybe make this clear this is just supported Webhook networks and doesn't include unsupported networks like Solana / Astar yet? 

Unless it's a 1:1 mapping to the SDK Network object, in which case it would probably make more sense to just skip the mapping and use the original object

deric-alchemy (approved):

src/api/notify-namespace.ts
thebrianchen: Great suggestion, much cleaner!

src/api/notify-namespace.ts
thebrianchen: I originally had it that way, but typescript can't recognize the overloads for some reason and will throw errors since it thinks that `params` can be any of the different param types. This then requires a cast every time we access params. Not sure if it's because I'm missing something.

src/api/notify-namespace.ts
thebrianchen: Good point, added a sentence at the top of `NotifyNamespace`. It's not a 1:1 mapping, but we need the mappings here internally to convert the SDK's `Network` num into the string format that the Notify API requires (and vice versa when creating the `Webhook` object).

src/api/notify-namespace.ts
dphilipson: Oh huh, you're right. I didn't realize that, good thinking!

dphilipson (approved):
Looks great!

src/internal/raw-interfaces.ts
```diff
@@ -208,3 +208,50 @@ export interface RawReingestContractResponse {
   reingestionState: string;
   progress: string | null;
 }
+
+export interface RawWebHook {
```
oceja, line 212: nit: is there a reason for the RawWebHook casing instead of RawWebhook?

src/api/notify-namespace.ts
```diff
@@ -0,0 +1,529 @@
...
+ * the documentation for which networks are supported.
+ *
+ * Do not call this constructor directly. Instead, instantiate an Alchemy object
+ * with `const alchemy = new Alchemy(config)` and then access the notify
+ * namespace via `alchemy.notify`.
+ */
+export class NotifyNamespace {
+  constructor(private readonly config: AlchemyConfig) {}
+
+  /** Get all webhooks from every app on your team. */
```
oceja, line 53: Nit: Since webhooks are not always tied with apps we might want to rename to `Get all webhooks on your team.`

src/types/types.ts
```diff
@@ -1099,3 +1105,223 @@ export enum TransactionJobStatus {
...
+ * A Dropped Transaction webhook is used to notify your app whenever a
+ * transaction sent through your API key gets dropped. This can be useful if you
+ * want to notify customers that their transactions were dropped.
+ */
+export interface DroppedTransactionWebhook extends Webhook {
+  type: WebhookType.DROPPED_TRANSACTION;
+}
+
+/**
+ * An Address Activity Webhook tracks ETH, ERC-20, and ERC721 transfers for the
```
oceja, line 1167: And ERC1155

src/types/types.ts
```diff
@@ -1099,3 +1105,223 @@ export enum TransactionJobStatus {
...
+ * provided addresses. This can be used to notify your app with real-time state
+ * changes when your tracked addresses send or receive tokens.
+ */
+export interface AddressActivityWebhook extends Webhook {
+  type: WebhookType.ADDRESS_ACTIVITY;
+}
+
+/**
+ * The NFT Activity Webhook tracks all ERC821 and ERC1155 activity. This can be
```
oceja, line 1176: *ERC721

oceja (approved):
🪝

thebrianchen (commented):
thanks for the detailed review and catching the typos omar!

Issue #140, opened 2022-10-06
FR: add support for newly released NFT APIs


thebrianchen:
Tracking issue for several new NFT endpoints that need to be added into the SDK:

- [X] [computeRarity](https://docs.alchemy.com/reference/computerarity): Done in #148 
- [X] [searchContractMetadata](https://docs.alchemy.com/reference/searchcontractmetadata) Done in #163
- [X] [summarizeNFTAttributes](https://docs.alchemy.com/reference/summarizenftattributes) Done in #155 
- [X] [getNftSales](https://docs.alchemy.com/reference/getnftsales): Done in #190 
- [X] [reportSpamContract](https://docs.alchemy.com/reference/reportspamcontract) (Not relevant to SDK)
- [X] [getNftMetadataBatch](https://docs.alchemy.com/reference/getnftmetadatabatch): Done in #189
- [X] [getContractsForOwner](https://docs.alchemy.com/reference/getcontractsforowner): Done in #201

xeno097:
Hi, I'd like to work on some of those missing NFT endpoints. I'd start with the `computeRarity` one and then maybe implement the others. Are there any guidelines I should follow to implement the feature?

thebrianchen:
@xeno097 Thanks for offering to help! I'll work on a more holistic contributing document in the coming weeks.

For the NFT APIs: 
- Add the top-level method into the `NftNamespace` and the underlying implementation in `src/internal/nft-api.ts`.
- Add unit tests in `test/unit/nft-api.test.ts`.
- Add a couple integration tests into `test/integration/nft.test.ts`.

Thanks! Just a heads up, we're working on changing the payloads in `getNftSales`, so some of the API surface area might change.

xeno097:
Nice! I'll try to get something done over the weekend and keep an eye on the repo for the changes.

xeno097:
@thebrianchen do you prefer that I keep working with smaller prs like the last one or a big one with everything? I'd prefer working with smaller ones because I think is tidier.

thebrianchen:
Prefer smaller PRs as well! Easier to manage and review for sure.

PR #148, opened 2022-10-16
Add Nft Api Compute Rarity

xeno097:
### Changelog

- Adds the `computeRarity` method in the `nft` namespace.
- Adds the `NftAttributeRarity` type to represent the result of the `computeRarity` endpoint.
- Fixes a typo in the `OwnedBaseNftsResponse` type doc.
- Adds unit and integration tests for the `computeRarity` method.

### Extra
I was considering wrapping the `computeRarity` endpoint result in an object with the following structure `{attributes: NftAttributeRarity[] }` and exposing that in the public API but then decided to keep the same format as the endpoint response. Also, I'll gladly accept suggestions for the `NftAttributeRarity` type name if you have a better name for this use case.
The other endpoints listed in #140 will be added in another pr.


src/types/types.ts
```diff
@@ -760,6 +760,22 @@ export interface GetFloorPriceResponse {
...
+ * Information about the rarity of an NFT's attribute in the specified collection.
+ *
+ * @public
+ */
+export interface NftAttributeRarity {
+  /** Name of the NFT's attribute. */
+  value: string;
+
+  /** The type of NFT attribute. */
+  trait_type: string;
```
thebrianchen, line 773: Please rename to `traitType` in order to be consistent with JS style.

src/types/types.ts
```diff
@@ -760,6 +760,22 @@ export interface GetFloorPriceResponse {
...
+
+  /** The NFT's attribute frequency in the current collection. */
```
thebrianchen, line 775: Please update comment to:
`A number from 0 to 1 representing the prevalence of this value for this trait type in the current collection.`

test/integration/nft.test.ts
```diff
@@ -240,6 +240,16 @@ describe('E2E integration tests', () => {
...
   });
 
+  it('computeRarity()', async () => {
+    const contractAddress = '0x0510745d2ca36729bed35c818527c4485912d99e';
+    const tokenId = '403';
+
+    const response = await alchemy.nft.computeRarity(contractAddress, tokenId);
+
+    expect(response).toBeDefined();
+    expect(response.length).toBeGreaterThan(0);
```
thebrianchen, line 250: Please update the test to check that each field in the response type is present in the first element of the array.

test/unit/nft-api.test.ts
```diff
@@ -1203,6 +1204,47 @@ describe('NFT module', () => {
...
+
+    it('calls with the correct parameters', async () => {
+      await alchemy.nft.computeRarity(contractAddress, tokenId);
+
+      expect(mock.history.get.length).toEqual(1);
+      expect(mock.history.get[0].params).toHaveProperty(
+        'contractAddress',
+        contractAddress
+      );
+      expect(mock.history.get[0].params).toHaveProperty('tokenId', tokenId);
```
thebrianchen, line 1235: Once you update `trait_type` to `traitType`, can you also add a test to check that the method converts the field properly?

thebrianchen (changes requested):
Thanks for taking this on -- we really appreciate it! Also thanks for catching some typos :)

Left a few nits and comments, but this is really good 🔥

src/types/types.ts
xeno097: On it 👌🏼

src/types/types.ts
xeno097: Okis, much better

test/unit/nft-api.test.ts
xeno097: Sure, no problem.

test/integration/nft.test.ts
xeno097: Thank you for the suggestion!

test/unit/nft-api.test.ts
```diff
@@ -1204,6 +1209,54 @@ describe('NFT module', () => {
     });
   });
 
+  describe('computeRarity()', () => {
+    const contractAddress = '0xbc4ca0eda7647a8ab7c2061c2e118a18a936f13d';
+    const tokenId = '7495';
```
thebrianchen, line 1214: one more request: can you use a numerical value for `tokenId` here to check that the logic to convert `Bignumberish` input to a string is properly executed? Thanks!

thebrianchen (approved):
Thanks for the clean PR and fast turnaround times! Left one last request, but will merge this in after that 🚀

test/unit/nft-api.test.ts
xeno097: Yup!

xeno097:
> Thanks for the clean PR and fast turnaround times! Left one last request, but will merge this in after that rocket

🔥

test/unit/nft-api.test.ts
xeno097: Ready 👌🏼

PR #152, opened 2022-10-18
Update gas optimized tx to new endpoints

thebrianchen:
This PR updates the endpoints and naming to use the new endpoints. I made some revisions to the method names as well. Keeping these public for testing, but the `@internal` tag tells rollup not to include it in the final bundle. Will update that once we are ready to launch.


deric-alchemy (approved):
Looks good!

PR #154, opened 2022-10-19
Add alchemy_minedTransactions

thebrianchen:
This PR adds support for the `alchemy_minedTransactions` endpoint in the Subscriptions API.

This requires some refactoring and hacking around with the ethers.js event tags. Some changes:
- Moved all logic related to Ethers.js events to a separate `ethers-event.ts` file.
- Added logic to serialize/deserialize the new subscription event. See the unit test `verifyRoundTrip()` calls to see how the objects are serialized.
- Consolidated event filter methods to `AlchemySubscription` enums

@dphilipson Would love feedback on how to keep the ethers event method clean. The `EthersEvent` class needs to support all of the different subscription types that we eventually add. However, when the properties on events overlap, we run into a serialization issue, since the serialization method is based on the property ordering in the object. In this case, `hashesOnly` is the 3rd property on `alchemy_pendingTransactions`, and it luckly happens to be the 3rd property in `alchemy_minedTransactions` as well. I'm wondering if there's a more graceful implementation that doesn't involve creating separate methods for each one.


src/internal/ethers-event.ts
```diff
@@ -0,0 +1,322 @@
...
+    } else {
+      return undefined;
+    }
+  }
+
+  /**
+   * Converts the event tag into the original `hashesOnly` field in
+   * {@link AlchemyPendingTransactionsEventFilter} and {@link AlchemyMinedTransactionsEventFilter}.
+   */
+  get hashesOnly(): boolean | undefined {
```
thebrianchen, line 66: This code is the tricky part I was referring to in the PR description since it relies on the the `hashesOnly` property being serialized in the 3rd position on both event types.

src/internal/ethers-event.ts
```diff
@@ -0,0 +1,322 @@
...
+ *
+ * The serialization and deserialization mechanism requires the order of the
+ * fields to be the same across different {@link ALCHEMY_EVENT_TYPES}. Before
+ * using a getter on a new event filter, make sure that the position of the
+ * field in serialization is correct.
+ *
+ * The getters on this class deserialize the event tag generated by
+ * {@link getAlchemyEventTag} into the original fields passed into the event.
+ */
+export class EthersEvent extends Event {
```
dphilipson, line 29: Pretty interesting! What you have here is fine since there's only a few event types so it's easy to keep track of all of them.

If we were to have a lot of event types, we'd probably want to define a "schema" for each event type that might look something like
```ts
const EVENT_TYPE_TO_SCHEMA = {
  [ALCHEMY_PENDING_TRANSACTIONS_EVENT_TYPE]:
    [
      ["fromAddress": StringParam],
      ["toAddress": StringParam],
      ["hashesOnly": BooleanParam],
    ],
  [ALCHEMY_MINED_TRANSACTIONS_EVENT_TYPE]:
    [
      ["addresses", AddressesParam],
      ["includeRemoved", BooleanParam],
      ["hashesOnly", BooleanParam],
    ],
}
```
where each of those `Param` objects contains the logic of how to serialize and deserialize a field of that type. Then you might use the above in a helper `deserialize` that acts like
```ts
const fields = deserialize(eventTag)
// returns e.g.:
//  {
//     "method": "alchemy_pendingTransactions",
//     "fromAddress": "0xABC",
//     "toAddress": ["0xDEF", "0xGHI"],
//     "hashesOnly: true
//   }
Object.assign(this, fields);
```
But that's a ton more work and we can worry about that if we end up with a bunch of event types later on.

dphilipson (approved):
Looks good, great testing. Fun to think about what the cleanest way to express this sort of thing is. I have an idea in the comment below, but it might be overkill for what we have at the moment.

noam-alchemy (approved):

src/internal/ethers-event.ts
thebrianchen: Thanks! Will keep this in mind as we add more endpoints.

PR #155, opened 2022-10-20
Nft api summarize nft attributes

xeno097:
### Changelog

- Adds  the `summarizeNft` method in the `nft` namespace.
- Adds the `NftAttributesSummary` type to describe the result for the `summarizeNft` endpoint.
- Adds unit and integration tests for the `summarizeNft` method.


src/internal/nft-api.ts
```diff
@@ -378,6 +379,19 @@ export async function computeRarity(
   return getNftRarityFromRaw(response);
 }
 
+export async function summarizeNftAttributes(
+  config: AlchemyConfig,
+  contractAddress: string,
+  srcMethod = 'summarizeNFTAttributes'
```
thebrianchen, line 385: can you rename to `summarizeNftAttributes`? the goal is to have the original name of the sdk method.

src/internal/nft-api.ts
```diff
@@ -378,6 +379,19 @@ export async function computeRarity(
...
+): Promise<NftAttributesSummary> {
```
thebrianchen, line 386: please rename `NftAttributesSummary` to `NftAttributesResponse`.

test/unit/nft-api.test.ts
```diff
@@ -1260,6 +1261,45 @@ describe('NFT module', () => {
...
+    it('calls with the correct parameters', async () => {
+      mock.onGet().reply(200, templateResponse);
+
+      await alchemy.nft.summarizeNftAttributes(contractAddress);
+
+      expect(mock.history.get.length).toEqual(1);
+      expect(mock.history.get[0].params).toHaveProperty(
+        'contractAddress',
+        contractAddress
+      );
```
thebrianchen, line 1290: can you add a few checks to verify that the response contains all the properties?

thebrianchen (changes requested):
thanks for taking another one on! just a few stylistic nits. excited to get this shipped!

src/types/types.ts
```diff
@@ -781,6 +781,22 @@ export interface NftAttributeRarity {
...
+ */
+export interface NftAttributesSummary {
+  /** The current NFT collection address. */
+  contractAddress: string;
+
+  /** The current NFT collection total supply. */
+  totalSupply: number;
+
+  /** The current NFT collection attribute prevalence summary grouped by trait type. */
+  summary: Record<string, Record<string, number>>;
```
thebrianchen, line 797: please modify comment to:
`The attribute prevalence of each trait grouped by the trait type for the provided NFT.`

src/internal/nft-api.ts
xeno097: Yup!

src/internal/nft-api.ts
xeno097: Oky

test/unit/nft-api.test.ts
xeno097: Sure 👍🏼

src/types/types.ts
xeno097: Considering that the address is for an NFT collection, should I add `collection` at the end of the prop description? The final text would be: `The attribute prevalence of each trait grouped by the trait type for the provided NFT collection`. I'll upload it as requested for the moment.

thebrianchen: Thanks! So to clarify, we want to standardize our wording to always to use NFT contract. Opensea has the concept of an "NFT collection" that can contain multiple different NFT contracts within it. So to avoid naming confusion, we want to clarify that this is for the individual contract, rather than the entire collection.

thebrianchen (approved):
woohoo! thanks for another contribution @xeno097 🔥

src/types/types.ts
xeno097: Got it, thank you for the clarification! 🔥

PR #157, opened 2022-10-20
Rename getAll() to getAllWebhooks()

thebrianchen:
Renaming based on discussion. Also removing the `webhooks:` prefix since the CRUD logs live in a separate imply cluster.


src/api/notify-namespace.ts
```diff
@@ -211,15 +211,15 @@ export class NotifyNamespace {
     let data;
     if ('isActive' in update) {
       restApiName = 'update-webhook';
-      methodName = 'webhooks:updateWebhook';
```
deric-alchemy, line 214: perhaps leave the webhooks prefix in the case that non-webhook CRUD fields get added to the CRUD logs?

deric-alchemy (approved):

src/api/notify-namespace.ts
thebrianchen: only the SDK sends the header, so it should be fine. I don't foresee a method name collision even if we add non-crud endpoints.

PR #163, opened 2022-10-28
Nft api search contract metadata

xeno097:
### Changelog

- Adds the `searchContractMetadata` method in the `nft` namespace.
- Refactors unit tests for the `getNftContractMetadata` method.
- Moves the `verifyNftContractMetadata` util function to the `test/test-util.ts` file to reuse it for the `searchContractMetadata` unit tests.
- Adds unit and integration tests for `searchContractMetadata` method.

### DIFF
--- a/test/test-util.ts
+++ b/test/test-util.ts
@@ -8,7 +8,8 @@ import {
   OwnedBaseNft,
   OwnedNft,
   TokenUri,
-  toHex
+  toHex,
+  NftContract
 } from '../src';
 import {
   RawBaseNft,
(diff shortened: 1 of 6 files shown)


test/test-util.ts
```diff
@@ -203,6 +204,23 @@ export function createRawNftContractBaseNft(
   };
 }
 
+export function verifyNftContractMetadata(
+  actualNftContract: NftContract,
+  expectedNftContract: NftContract,
+  address: string,
```
thebrianchen, line 210: why do we need all the additional variables in the params here?

i'm thinking that we would just do
`expect(actualNftContract.address).toEqual(expectedNftContract.address);`

thebrianchen (changes requested):
thanks for submitting this! just a small change and we should be good to go!

test/test-util.ts
xeno097: Playing with the API I saw that every element in the `searchContractMetadata` endpoint response has the same structure as the response from the `getNftContractMetadata` endpoint which was unit tested using the following function 

image
image

So to keep the testing method consistent I decided to reuse the `verifyNftContractMetadata` function for the `searchContractMetadata` tests too.

thebrianchen: ah I see what you mean, thanks for clarifying!

please leave this as is, i'll refactor this later!

xeno097: Ok, nice! Thank you!

thebrianchen (approved):
thanks again!

PR #165, opened 2022-10-30
NFT Unichain API for getNftsForOwners

brg8:
## Motivation

Some developers have expressed that it is annoying to call NFT API endpoints for each network that they are interested in. Some end-users may want to view their NFTs across all networks at once, for instance. This PR attempts to address this feedback for the `getNftsForOwner` method.

## Request Ergonomics

In order to keep requests as ergonomic as possible, I decided to add the unichain api to the existing NFT namespace. That means developers will still initialize an `Alchemy` class with an initial network. This feels a little strange, as the developer using the unichain API will still be using a config that has a single network in it. However, it felt preferrable to other options, e.g. creating a new `Network` type called "Unichain".

The unichain API takes one additional parameter to the regular API, which is an array of `networks: Network[]`. The `options` dictionary is still supported and will be passed into each call to `getNftsForOwner`.

## Response Format

At first I considered keeping the response object for the unichain call the same as for `getNftsForOwner`, i.e.

```
export interface OwnedNftsResponse {
  /** The NFTs owned by the provided address. */
  readonly ownedNfts: OwnedNft[];
  /**
   * Pagination token that can be passed into another request to fetch the next
   * NFTs. If there is no page key, then there are no more NFTs to fetch.
   */
  readonly pageKey?: string;
  /** The total count of NFTs owned by the provided address. */
  readonly totalCount: number;
}
```

However, a single list of owned NFTs would not be distinguishable by network. I would have to create a new `OwnedNft` type with a network field.

Another (perhaps larger) reason to create a new response is that pagination and sorting are an open question when it comes to this response format. Since we are making requests to multiple networks we will usually have more than 100 results. How do we decide which 100 to return? How do we sort them? How do we handle pagination? I decided to remove this complexity for now.

## Implementation

The concept behind the unichain API is simple -- the developer provides a set of networks and we call `getNftsForOwner` for each network. The complicated part is managing page keys.

In a single-network API (read: all current APIs), the developer knows they have reached the end of the result set when the API does _not_ return a page key. For a unichain request, however, some networks may be out of pages while others have many more to go. This makes the unichain page key a bit of a sticky wicket. The one behavior I sought to maintain is that if the unichain API does _not_ return a page key that means the developer is done with _all_ pages on _all_ networks.

In any case... happy reviewing!

### DIFF
--- a/src/internal/nft-api-unichain.ts
+++ b/src/internal/nft-api-unichain.ts
@@ -12,7 +12,7 @@ import { getNftsForOwner } from './nft-api';
 import { NetworkPageKey, UnichainPageKeyCache } from './page-key';
 
 function networksAreEqual(a: Network[], b: Network[]): boolean {
-  return JSON.stringify(a) === JSON.stringify(b);
+  return JSON.stringify(a.sort()) === JSON.stringify(b.sort());
 }
 
 function validateParams(
--- a/src/internal/nft-api-unichain.ts
+++ b/src/internal/nft-api-unichain.ts
@@ -1,0 +1,155 @@
+import { AlchemyConfig } from '../api/alchemy-config';
+import {
+  GetBaseNftsForOwnerUnichainOptions,
+  GetNftsForOwnerUnichainOptions,
+  Network,
+  OwnedBaseNftsResponse,
+  OwnedBaseNftsResponseUnichain,
+  OwnedNftsResponse,
+  OwnedNftsResponseUnichain
+} from '../types/types';
+import { getNftsForOwner } from './nft-api';
+import { NetworkPageKey, UnichainPageKeyCache } from './page-key';
+
+function networksAreEqual(a: Network[], b: Network[]): boolean {
+  return JSON.stringify(a) === JSON.stringify(b);
+}
+
+function validateParams(
+  networks: Network[],
+  options:
+    | GetNftsForOwnerUnichainOptions
+    | GetBaseNftsForOwnerUnichainOptions = {}
+): {
+  unichainPageKeyCache: UnichainPageKeyCache;
+  networkPageKeys: Map<Network, NetworkPageKey>;
+} {
+  // We could default to all networks, but as we add or remove support for
+  // various networks that list can change. We force developers to explicitly
+  // list their desired networks so that their code is future-proof.
+  if (!networks || networks.length === 0) {
+    throw new Error(
+      'Must provide a `networks` parameter for unichain requests.'
+    );
+  }
+
+  const { unichainPageKeyCache, pageKey = null } = options;
+
+  // This should be provided by default in the namespace, so realistically
+  // should never happen. In non-unichain methods we do not use a page key
+  // cache because page keys are handled by the API. However we have no API
+  // for unichain, and so the SDK needs to handle page keys.
+  if (!unichainPageKeyCache) {
+    throw new Error('No page key cache was provided.');
+  }
+
+  const networkPageKeys = unichainPageKeyCache.getNetworkPageKeys(pageKey);
+
+  // If the user made an initial request for a set of networks, we do not
+  // allow them to change networks on the next page. We could technically
+  // allow this, but it would require significantly more magic.
+  const originalNetworks = Array.from(networkPageKeys.keys());
+  if (pageKey && !networksAreEqual(originalNetworks, networks)) {
+    throw new Error('Networks must not change between pages.');
+  }
+
+  return {
+    unichainPageKeyCache,
+    networkPageKeys
+  };
+}
+
+function buildResponseWithPageKey(
+  networks: Network[],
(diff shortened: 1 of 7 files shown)


src/internal/nft-api-unichain.ts
```diff
@@ -0,0 +1,155 @@
...
+  OwnedBaseNftsResponse,
+  OwnedBaseNftsResponseUnichain,
+  OwnedNftsResponse,
+  OwnedNftsResponseUnichain
+} from '../types/types';
+import { getNftsForOwner } from './nft-api';
+import { NetworkPageKey, UnichainPageKeyCache } from './page-key';
+
+function networksAreEqual(a: Network[], b: Network[]): boolean {
+  return JSON.stringify(a.sort()) === JSON.stringify(b.sort());
```
dphilipson, line 15: Better to not mutate the input. One way to do this would be to use `[...a].sort()`, etc. instead.

src/internal/nft-api-unichain.ts
```diff
@@ -0,0 +1,155 @@
...
+
+function buildResponseWithPageKey(
+  networks: Network[],
+  networkResults: (OwnedNftsResponse | OwnedBaseNftsResponse)[],
+  unichainPageKeyCache: UnichainPageKeyCache
+) {
+  const response: OwnedNftsResponseUnichain = {
+    nftsByNetwork: new Map()
+  };
+  const networkPageKeys = new Map();
```
dphilipson, line 70: Should help out the type system and readers a little here, either
```ts
const networkPageKeys: NetworkPageKeys = new Map();
```
or
```ts
const networkPageKeys = new Map<Network, NetworkPageKey>();
```
Otherwise, the type system can't infer what kind of `Map` this is.

src/internal/page-key.ts
```diff
@@ -0,0 +1,89 @@
...
+interface UnichainPageKeyValue {
+  networkPageKeys: NetworkPageKeys;
+  expiresAt: number;
+}
+
+export class NetworkPageKey {
+  // When a network has no next page we will save a page key with this
+  // value. The hasNextPage() method will return false for any page key
+  // matching this value.
+  private NO_MORE_PAGES = 'NO_MORE_PAGES';
```
dphilipson, line 14: Can be `static` and `readonly`.

src/internal/page-key.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  private pageKey: string;
```
dphilipson, line 16: Can be `readonly`.

src/internal/page-key.ts
```diff
@@ -0,0 +1,89 @@
...
+  private pageKeyExpirationMs: number;
+  private cache: Record<string, UnichainPageKeyValue>;
+
+  constructor(
+    cache: Record<string, UnichainPageKeyValue> = {},
+    pageKeyExpirationMs: number = DEFAULT_UNICHAIN_PAGE_KEY_EXPIRATION_MS
+  ) {
+    this.cache = cache;
+    this.pageKeyExpirationMs = pageKeyExpirationMs;
+  }
```
dphilipson, line 46: This can be written more simply as
```ts
export class UnichainPageKeyCache {
  // (No field declarations here)

  constructor(
    private readonly cache: Record<String, UnichainPageKeyValue> = {},
    private readonly pageKeyExpirationMs = DEFAULT_UNICHAIN_PAGE_KEY_EXPIRATION_MS,
  ) {}
  
  // ... rest of class
}
```

src/types/types.ts
```diff
@@ -542,6 +543,24 @@ export interface GetNftsForOwnerOptions {
...
+export interface GetNftsForOwnerUnichainOptions extends GetNftsForOwnerOptions {
+  /**
+   * A local cache with a map of unichain pageKey to network pageKeys.
+   */
+  unichainPageKeyCache?: UnichainPageKeyCache;
+
+  /**
+   * Function to call to get an owner's nfts for a single network.
+   */
+  getNftsForOwnerFn?: Function;
```
dphilipson, line 555: There must be a more specific type than this, likewise below.

dphilipson (commented):
Hello, just doing a drive-by code review about pedantic stuff! Still need to think over a bit about the behavior here.

src/internal/page-key.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  public set(unichainKey: string, networkPageKeys: NetworkPageKeys): void {
+    this.cache[unichainKey] = {
+      networkPageKeys,
+      expiresAt: Date.now() + this.pageKeyExpirationMs
+    };
+  }
+
+  public get(unichainKey: string): UnichainPageKeyValue | null {
+    return this.cache[unichainKey] || null;
```
dphilipson, line 67: This is fine, but I like to use `??` instead of `||` when expressing default values like this. It does the same thing, but only uses the default value if the left-hand-side is `null` or `undefined`, rather than any falsy value.

Likewise in test file.

PR #170, opened 2022-11-03
Add uniqueId for AssetTransfersResult

CocoJr:
Just figure out that the `uniqueId` field is missing from the type.


thebrianchen (approved):
you're right, thanks for catching this!

PR #184, opened 2022-11-09
Support ENS resolution in WebsocketNamespace

thebrianchen:
To support ENS resolution in the WebsocketNamespace, we need to resolve ENS names at the namespace level before passing down the original event with resolved names down to the underlying websocket provider.

Also added some unit tests to verify that the resolution logic works, but opted out of adding tests for the top-level namespace methods.


deric-alchemy (approved):

dphilipson (approved):
Great tests as always!

PR #185, opened 2022-11-10
Add support for automatic batching in json-rpc requests

### LINKED ISSUE #159: FR: Add support for batched requests
Implement JsonRpcBatchProvider into the existing provider to facilitate making batch requests. This will unblock adding syntactic sugar on top of existing methods to make it easier to combine responses from different endpoints into a single one.

thebrianchen:
Fixes #159.

This PR adds support for automatic batching of requests based loosely on ethers.js's [JsonRpcBatchProvider](https://github.com/ethers-io/ethers.js/blob/ce8f1e4015c0f27bf178238770b1325136e3351a/packages/providers/src.ts/json-rpc-batch-provider.ts). The main logic is encapsulated in the `RequestBatcher` class and we additionally enforce a max batch size of 100.

### DIFF
--- a/src/api/alchemy-provider.ts
+++ b/src/api/alchemy-provider.ts
@@ -40,7 +40,7 @@ export class AlchemyProvider
 {
   readonly apiKey: string;
   readonly maxRetries: number;
-  readonly optimizedBatching: boolean;
+  readonly batchRequests: boolean;
 
   /**
    * VISIBLE ONLY FOR TESTING
@@ -77,7 +77,7 @@ export class AlchemyProvider
 
     this.apiKey = config.apiKey;
     this.maxRetries = config.maxRetries;
-    this.optimizedBatching = config.optimizedBatching;
+    this.batchRequests = config.batchRequests;
 
     // TODO: support individual headers when calling batch
     const batcherConnection = { ...this.connection };
@@ -250,7 +250,7 @@ export class AlchemyProvider
     const connection = { ...this.connection };
     connection.headers!['Alchemy-Ethers-Sdk-Method'] = methodName;
 
-    if (this.optimizedBatching || forceBatch) {
+    if (this.batchRequests || forceBatch) {
       return this.batcher.enqueueRequest(request as JsonRpcRequest);
     }
     // END MODIFIED CODE
--- a/src/api/alchemy-provider.ts
+++ b/src/api/alchemy-provider.ts
@@ -10,6 +10,8 @@ import {
 } from '@ethersproject/providers';
 import { ConnectionInfo, fetchJson } from '@ethersproject/web';
 
+import { JsonRpcRequest } from '../internal/internal-types';
+import { RequestBatcher } from '../internal/request-batcher';
 import { Network } from '../types/types';
 import {
   CustomNetworks,
@@ -38,6 +40,11 @@ export class AlchemyProvider
 {
   readonly apiKey: string;
   readonly maxRetries: number;
+  readonly optimizedBatching: boolean;
+
+  /** VISIBLE ONLY FOR TESTING
+   *@internal */
+  readonly batcher: RequestBatcher;
 
   /** @internal */
   constructor(config: AlchemyConfig) {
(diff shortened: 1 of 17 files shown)


src/internal/request-batcher.ts
```diff
@@ -0,0 +1,126 @@
...
+    });
+
+    this.pendingBatch.push(inflightRequest);
+
+    if (this.pendingBatch.length === this.maxBatchSize) {
+      // Send batch immediately if we are at the maximum batch size.
+      void this.sendBatchRequest();
+    } else if (!this.pendingBatchTimer) {
+      // Schedule batch for next event loop + short duration
+      this.pendingBatchTimer = setTimeout(() => this.sendBatchRequest(), 10);
```
deric-alchemy, line 60: made the "10" a global variable so event loop duration is more visible?

src/types/types.ts
```diff
@@ -40,6 +40,14 @@ export interface AlchemySettings {
...
    */
   authToken?: string;
+
+  /**
+   * Optional setting that automatically batches and sends json-rpc requests for higher
+   * throughput and reduced network IO. Defaults to false.
+   *
+   * This implementation is based on the `JsonRpcBatchProvider` in ethers.
+   */
+  optimizedBatching?: boolean;
```
deric-alchemy, line 50: maybe rename this to `batchRequests`? Unless it was previously named this in Ethers.js or something

deric-alchemy (approved):

src/api/alchemy-provider.ts
```diff
@@ -218,14 +233,28 @@ export class AlchemyProvider
...
+    forceBatch = false
+  ): Promise<any> {
     const request = {
       method,
       params,
       id: this._nextId++,
       jsonrpc: '2.0'
     };
 
+    // START MODIFIED CODE
```
dphilipson, line 249: Thanks for having these comments!

src/types/types.ts
thebrianchen: done.

src/internal/request-batcher.ts
```diff
@@ -0,0 +1,126 @@
...
+   * Enqueues the provided request. The batch is immediately sent if the maximum
+   * batch size is reached. Otherwise, the request is enqueued onto a batch that
+   * is sent after 10ms.
+   *
+   * Returns a promise that resolves with the result of the request.
+   */
+  async enqueueRequest(request: JsonRpcRequest): Promise<any> {
+    if (this.pendingBatch === undefined) {
+      this.pendingBatch = [];
+    }
```
dphilipson, line 45: (Minor) The logic might be simpler if you just have `pendingBatch` be an empty array instead of `undefined` in the case where it's empty. Everything would work out the same (with you checking for 0 length instead of undefined), except that you wouldn't need this one if-block here.

src/internal/request-batcher.ts
```diff
@@ -0,0 +1,126 @@
...
+    // batch (or vice-versa).
+    if (this.pendingBatch === undefined) {
+      return;
+    }
+
+    // Get the current batch and clear it, so new requests
+    // go into the next batch
+    const batch = this.pendingBatch!;
+    this.pendingBatch = undefined;
+    this.pendingBatchTimer = undefined;
```
dphilipson, line 90: You might want to `clearTimeout` on the timer here. That way if the batch send is triggered by the size limit, then it won't try to send a partial batch until a full 10 ms have passed.

src/internal/request-batcher.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * Array of enqueued requests along with the constructed promise handlers for
+   * each request.
+   */
+  private pendingBatch: Array<BatchRequest> | undefined;
+
+  constructor(
+    private readonly sendBatchFn: SendBatchFn,
+    private readonly connection: ConnectionInfo,
```
dphilipson, line 31: Instead of having this class take a complicated `SendBatchFn` as well as `ConnectionInfo`, I'd consider just having one param which is `(req: JsonRpcRequest) => Promise<JsonRpcResponse>`, and then the caller can be responsible for crafting that function to pass in. My reasoning is that passing the connection info around, serializing JSON, and converting the response are not related to batching at all. Further, it's likely that the class calling this (the provider) also needs this logic for the nonbatched case, and so as written you get some code duplication between this class and the caller.

dphilipson (commented):
Looks great! Nice work on this. Some suggestions inline.

src/internal/request-batcher.ts
```diff
@@ -68,23 +69,19 @@ export class RequestBatcher {
...
-    // This if-statement handles the case where the maximum batch size triggers
-    // the batch send, so the scheduled timeout send is called with an empty
-    // batch (or vice-versa).
-    if (this.pendingBatch === undefined) {
-      return;
-    }
-
     // Get the current batch and clear it, so new requests
     // go into the next batch
     const batch = this.pendingBatch!;
```
dphilipson, line 74: (Very nit) You don't need the `!` on `this.pendingBatch` anymore.

src/internal/request-batcher.ts
dphilipson: Oh whoops, I just realized I forgot to make the request and response arrays in the type I put in above. Thanks for figuring out what I meant to say!

dphilipson (approved):
Looks great, thanks for updating!

PR #189, opened 2022-11-11
Add getNftMetadataBatch()

thebrianchen:
This PR adds support for the `getNFTMetadataBatch` endpoint.

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -724,6 +724,35 @@ export interface NftContractTokenBalance {
 }
 
 /**
+ * Represents an NFT token to fetch metadata for in a {@link NftNamespace.getNftMetadataBatch}
+ * method.
+ */
+export interface NftMetadataBatchToken {
+  /** The NFT contract address. Limited to ERC721 and ERC1155 tokens. */
+  contractAddress: string;
+
+  /** The id of the NFT. */
+  tokenId: BigNumberish;
+}
+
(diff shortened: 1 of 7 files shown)


src/types/types.ts
```diff
@@ -723,6 +723,35 @@ export interface NftContractTokenBalance {
   balance: number;
 }
 
+/**
+ * Represents an NFT token to fetch metadata for in a {@link NftNamespace.getNftMetadataBatch}
+ * method.
+ */
+export interface NftMetadataBatchToken {
```
niveda-krish, line 730: Should we add tokenType here?

src/types/types.ts
```diff
@@ -723,6 +723,35 @@ export interface NftContractTokenBalance {
...
+   * If you want to only access the cache and not live fetch any metadata for
+   * cache misses then set this value to 0.
+   */
+  tokenUriTimeoutInMs?: number;
+
+  /**
+   * Whether to refresh the metadata for the given NFT token before returning
+   * the response. Defaults to false for faster response times.
+   */
+  refreshCache?: boolean;
```
niveda-krish, line 752: Should we add contractAddressHint here?

src/types/types.ts
thebrianchen: I'm going off of the params from the docs: https://docs.alchemy.com/reference/getnftmetadatabatch

ayang-rkjs (approved):

PR #193, opened 2022-11-16
First pass at the External Wallet

deric-alchemy:


deric-alchemy:
Should we try to combine ExternalWallet and Wallet into one object and just have overloads for the constructor?

src/api/alchemy-external-wallet.ts
```diff
@@ -0,0 +1,166 @@
...
+   * @param name Name of the ENS address
+   * @override
+   */
+  resolveName(name: string): Promise<string> {
+    return this.signer.resolveName(name);
+  }
+
+  /** Forwards any relevant request to the send method in the Provider */
+  send(method: string, params: any[]): Promise<any> {
+    return this.provider.send(method, params);
```
dphilipson, line 164: Did you want this to be
```ts
return this.alchemy.provider.send(method, params)
```
or something like that?

deric-alchemy: It actually needs to send the requests to the provider of the wallet, in order to properly handle queries like eth_requestAccounts! 

We could build in support for sending requests to the Alchemy Provider by having a method called sendAlchemy() or something...but in my head, if they really want the Alchemy Provider, they should just create a separate Alchemy object

deric-alchemy: Update: I updated the comments on both the ExternalWallet and send() to accurately reflect that it will just send requests to the ExternalProvider. We won't support the Alchemy object in this package at this time (though later on we might)

PR #199, opened 2022-11-17
Move BaseNft comments to method docs, not general docs.

deric-alchemy:
Reason: too niche for the general docs (which are quickly becoming too verbose).

### DIFF
--- a/src/api/nft.ts
+++ b/src/api/nft.ts
@@ -10,6 +10,10 @@ import {
 /**
  * Alchemy representation of a base NFT contract that doesn't contain metadata.
  *
+ * The BaseNftContract does not hold any metadata information and only contains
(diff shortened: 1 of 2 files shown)


src/api/nft.ts
```diff
@@ -51,6 +63,10 @@ export interface BaseNft {
 /**
  * Alchemy representation of an NFT.
  *
+ * The BaseNft object does not hold any metadata information and only contains
+ * the NFT contract and token ID. The Nft object additionally contains the NFT
```
thebrianchen, line 67: can we wrap The references to the objects with `{@link}` to make it easier to read?

thebrianchen (approved):
looks good!

PR #210, opened 2022-12-09
Fix broken newHeads backfill

dphilipson:
newHeads were failing to backfill because of a misuse of the return value of `sendBatch`. Fix this up.


thebrianchen (approved):
thanks for debugging through this!

PR #211, opened 2022-12-09
Fix reconnections for other subscription types

dphilipson:
Subscription reconnections were only working for `newHeads` and `logs` subscriptions because of a missing case for the other types. Now all subscription types will reconnect when the connection drops.


thebrianchen (approved):
thanks for looking into this

PR #227, opened 2022-12-30
Support for tokenId null when using update webhook NFT filters

leonacostaok:
## Problem
The SDK is not supporting properly the API `update-webhook-nft-filters`. When trying to add collection addresses with tokenId null (to listen activity for all token ids in the collection), it doesn't allow you. [API reference](https://docs.alchemy.com/reference/update-webhook-nft-filters)

## Solution
1. Make the variable tokenId optional in the type and its usage.


src/types/types.ts
```diff
@@ -1777,7 +1777,7 @@ export interface NftFilter {
   /** The contract address of the NFT. */
   contractAddress: string;
   /** The token id of the NFT. */
-  tokenId: BigNumberish;
+  tokenId?: BigNumberish;
```
thebrianchen, line 1780: Please add a comment that omitting this field results in tracking all activity for the provided contract address. Thanks!

thebrianchen (changes requested):
thanks for catching this and supplying a fix! small comment to add some documentation.

PR #228, opened 2022-12-31
Fixed the typo, should be "getNftsForContractIterator" instead of "getNftForContractIterator"

SahilAujla:
The name of the endpoint is `getNftsForContractIterator` instead of `getNftForContractIterator`.


thebrianchen (approved):
thanks for catching!

PR #240, opened 2023-01-12
Fixed excludeZeroValue docs to be the same as API

j05u3:
Fixed alchemy_getAssetTransfers excludeZeroValue docs to be the same as the API where excludeZeroValue is said to default to `true` ([link](https://docs.alchemy.com/reference/alchemy-getassettransfers)).


thebrianchen (approved):
thanks for catching this discrepancy!

Issue #302, opened 2023-04-10
Incompatible with moduleResolution: bundler


florrdv:
### [REQUIRED] Environment

- Browser version: N/A
- Alchemy SDK version: 2.7.0

### [REQUIRED] Describe the problem
The alchemy-sdk package seems to be incompatible with the new `"moduleResolution": "bundler"` option introduced in TypeScript 5.0, this can be resolved by adding `"types": "./dist/src/index.d.ts"` to the `"exports"` section in `package.json`.

#### How to reproduce:
This issue can be reproduced by importing the alchemy-sdk package (latest version) in a TypeScript project with `moduleResolution` set to `bundler`.

florrdv:
Hi @thebrianchen! Thank you so much for the swift response. 

It looks like you need to enable the strict compiler options for it to show up. I've put up a repro example [here](https://github.com/florrdv/alchemy-declarations-missing-repro), which can be ran using the start script. I believe it is important to use `tsc` as well.

thebrianchen:
@florrdv Thanks for the repro example! Was able to see the error and how your PR fixes it.

thebrianchen:
Closing the issue since the fix has been merged. Will release in the `v2.8.1` early next week.

PR #303, opened 2023-04-10
Fix incompatibility with moduleResolution: bundler

florrdv:
Solves #302


thebrianchen (approved):
Thanks for the proactive investigation and fix! 💯

PR #311, opened 2023-04-25
Fix getTransfersForX() bug

thebrianchen:
Fixed a bug with `getTransfersForOwner()` and `getTransfersForContract()` methods in the `NftNamespace`, where some NFTs would not be returned if the NFT was transferred multiple times.

`getNftMetdataBatch()` de-dupes tokens. This caused a bug where the if a batch contained the same token twice, the second token would not be hydrated with nft metadata.

Changes:
- Create a mapping of each token to the underlying NFT that is used to hydrate the transfer metadata
- Added an integration test to verify the behavior
- Deleted extra integration test that was a duplicate.

### DIFF
--- a/src/internal/nft-api.ts
+++ b/src/internal/nft-api.ts
@@ -798,10 +798,27 @@ async function getNftsForTransfers(
     config,
     metadataTransfers.map(transfer => transfer.token)
   );
-  const transferredNfts = nfts.map((nft, i) => ({
-    ...nft,
-    ...metadataTransfers[i].metadata
-  }));
+
+  // The same NFT can be transferred multiple times in the same transfers response.
+  // We want to return one NFT for each transfer, so we create a mapping for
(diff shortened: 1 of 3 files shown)


src/internal/nft-api.ts
```diff
@@ -798,10 +798,27 @@ async function getNftsForTransfers(
...
+    const key = `${nft.contract.address}-${BigNumber.from(
+      nft.tokenId
+    ).toString()}`;
+    nftsByTokenId.set(key, nft);
+  });
+
+  const transferredNfts = metadataTransfers.map(t => {
+    const key = `${t.token.contractAddress}-${BigNumber.from(
+      t.token.tokenId
+    ).toString()}`;
```
brg8, line 816: Is it always the case that the `nfts` list will contain every NFT in the `metadataTransfers` list?

brg8 (approved):

PR #314, opened 2023-05-01
[SDK][Account Abstraction] Add Bundler and User Operation Endpoints

rthomare:
# This PR

# Test Plan


src/api/bundler-namespace.ts
```diff
@@ -0,0 +1,114 @@
...
+} from '../types/types';
+import { AlchemyConfig } from './alchemy-config';
+
+/**
+ * The Bundler namespace contains methods used for interacting performing user operations and
+ * checking on the state of submitted operations.
+ *
+ * Do not call this constructor directly. Instead, instantiate an Alchemy object
+ * with `const alchemy = new Alchemy(config)` and then access the transact
+ * namespace via `alchemy.paymaster`.
```
thebrianchen, line 15: via `alchemy.bundler`

src/api/alchemy.ts
```diff
@@ -69,5 +76,7 @@ export class Alchemy {
     this.transact = new TransactNamespace(this.config);
     this.notify = new NotifyNamespace(this.config);
     this.debug = new DebugNamespace(this.config);
+    this.paymaster = new PaymasterNamespace(this.config);
```
thebrianchen, line 79: nit: remove the paymaster line

src/api/bundler-namespace.ts
```diff
@@ -0,0 +1,114 @@
...
+  constructor(private readonly config: AlchemyConfig) {}
+
+  /**
+   * Sends a User Operation
+   *
+   * Sends a user operation to the provided bundler client evm network, and
+   * returns a user operation hash on successful submission.
+   *
+   * @param userOperation - The user operation.
+   * @public
```
thebrianchen, line 28: instead of using `@public` here, use `@beta` here and throughout for now, which lets us do breaking changes on the method later.

src/api/bundler-namespace.ts
```diff
@@ -0,0 +1,114 @@
...
+      'eth_getUserOperationByHash',
+      [userOpHash],
+      'getUserOperationByHash'
+    );
+  }
+
+  /**
+   * Get Client Supported Entry Points
+   *
+   * Returns an array of the entry point addresses that are
```
thebrianchen, line 62: can you explain what an entry point address is // used for? where do you pass the return of this method into?

src/types/types.ts
```diff
@@ -2591,6 +2591,93 @@ export interface DebugPrestate {
   storage: Record<string, string>;
 }
 
+/*
+ * SDK Representation of a user operation and bundler types
```
thebrianchen, line 2595: since this type is going to live with a bunch of other types, can you explain this in the context of the bundler namespace. ex:
```
Representation of a submitted user operation returned by the {@link BundlerNamespace.getUserOperationByHash} 
```

src/types/types.ts
```diff
@@ -2591,6 +2591,93 @@ export interface DebugPrestate {
...
+ */
+export interface UserOperation {
```
thebrianchen, line 2597: can we have UserOperation extend UserOperationRequest? Should be good to use inheritance here since they are all interfaces.

src/types/types.ts
```diff
@@ -2591,6 +2591,93 @@ export interface DebugPrestate {
...
+  sender: string;
```
thebrianchen, line 2598: can you add a short comment for each field here and below?

src/types/types.ts
```diff
@@ -2591,6 +2591,93 @@ export interface DebugPrestate {
...
+  maxPriorityFeePerGas: number;
+  paymasterAndData: string;
+  signature: string;
+  entryPoint: string;
+  blockNumber: number;
+  blockHash: string;
+  transactionHash: string;
+}
+
+export interface UserOperationRequest {
```
thebrianchen, line 2615: add comments for the interfaces here and below pls. for string fields that represent numeric values, also add clarification if it's a hex or integer string.

src/types/types.ts
```diff
@@ -2591,6 +2591,93 @@ export interface DebugPrestate {
...
+  transactionHash: string;
+  from: string;
+  to: string;
+  cumulativeGasUsed: string;
+  gasUsed: string;
+  contractAddress: string;
+  logs: UserOperationReceiptLog[];
+}
+
+export interface UserOperationEstiamateGasResponse {
```
thebrianchen, line 2672: s/Estiamate/Estimate

src/types/types.ts
```diff
@@ -2591,6 +2591,93 @@ export interface DebugPrestate {
...
+
+export interface UserOperationRequest {
+  sender: string;
+  nonce: string;
+  initCode: string;
+  callData: string;
+  callGasLimit: number;
+  verificationGasLimit: number;
+  preVerificationGas: number;
+  maxFeePerGas: number;
```
thebrianchen, line 2623: wanted to double check, are all fields here required? Aren't fields like `maxFeePerGas` and `maxPriorityFeePerGas` usually optional?

src/types/types.ts
```diff
@@ -2591,6 +2591,93 @@ export interface DebugPrestate {
...
+  transactionHash: string;
+}
+
+export interface UserOperationRequest {
+  sender: string;
+  nonce: string;
+  initCode: string;
+  callData: string;
+  callGasLimit: number;
+  verificationGasLimit: number;
```
thebrianchen, line 2621: do any of the `number` fields here overflow? if so, we should represent them as BigNumber for now, and i'll replace it with BigInt for v6.

src/types/types.ts
```diff
@@ -2591,6 +2591,93 @@ export interface DebugPrestate {
...
+  preVerificationGas: number;
+  maxFeePerGas: number;
+  maxPriorityFeePerGas: number;
+  paymasterAndData: string;
+  signature: string;
+  entryPoint: string;
+  blockNumber: number;
```
thebrianchen, line 2610: just to clarify, is blockNumber a `number` here and `string` in other occurrences? is it possible to standardize?

src/types/types.ts
```diff
@@ -2591,6 +2591,93 @@ export interface DebugPrestate {
...
+  to: string;
+  cumulativeGasUsed: string;
+  gasUsed: string;
+  contractAddress: string;
+  logs: UserOperationReceiptLog[];
+}
+
+export interface UserOperationEstiamateGasResponse {
+  /* Gas overhead of this UserOperation */
+  preVerificationGas: number;
```
thebrianchen, line 2674: do gas fields have a chance of overflowing on different chains? i'm leaning that we standardize to returning the response as an integer string and casting to BigNumber (to be replaced by BigInt)

thebrianchen (commented):
Thanks for taking this on! left some nits and questions 🚀

src/types/types.ts
```diff
@@ -2591,6 +2591,93 @@ export interface DebugPrestate {
...
+export interface UserOperation {
+  sender: string;
+  nonce: string;
+  initCode: string;
+  callData: string;
+  callGasLimit: number;
+  verificationGasLimit: number;
+  preVerificationGas: number;
+  maxFeePerGas: number;
+  maxPriorityFeePerGas: number;
```
moldy530, line 2606: these need to be `bigint`, and are represented as `hex` when transmitted

src/types/types.ts
moldy530: +1, these should be `bigint` types

src/types/types.ts
rthomare: Hmm not according to our api: https://github.com/OMGWINNING/alchemy-api-executors/blob/3b6054e7/evm-service/src/main/java/alchemy/evmservice/methods/rundler/EthSendUserOperation.java#L28

rthomare: Let me ask @dancoombs and @dphilipson

rthomare: Looks like we do https://alchemyinsights.slack.com/archives/C04RHSDSP9U/p1683141735130749

src/types/types.ts
```diff
@@ -2591,6 +2591,93 @@ export interface DebugPrestate {
   storage: Record<string, string>;
 }
 
+/*
+ * SDK Representation of a user operation and bundler types
+ */
+export interface UserOperation {
+  sender: string;
+  nonce: string;
```
dphilipson, line 2599: The nonce is a `uint256`, so it should be a bigint or whatever too.

src/types/types.ts
```diff
@@ -2591,6 +2591,93 @@ export interface DebugPrestate {
...
+  verificationGasLimit: number;
+  preVerificationGas: number;
+  maxFeePerGas: number;
+  maxPriorityFeePerGas: number;
+  paymasterAndData: string;
+  signature: string;
+  entryPoint: string;
+  blockNumber: number;
+  blockHash: string;
+  transactionHash: string;
```
dphilipson, line 2612: What does this field mean?

src/types/types.ts
dphilipson: Does this type represent input that the user provides, or the actual data sent over the wire to the bundler? I'm assuming the former, since otherwise all the numbers should be hex strings. In that case, from the linked thread I would consider making all the gas fields optional, and if not provided have the sdk fill them in.

src/api/bundler-namespace.ts
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * Estimates the Gas Values for a UserOperation.
+   *
+   * This method returns gas estimates based on the provided user operation request.
+   *
+   * @param userOperation - The user operation.
+   * @public
+   */
+  async estimateUserOperationGas(
+    userOperation: UserOperationRequest
```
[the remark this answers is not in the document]
moldy530: yea something like this: https://github.com/OMGWINNING/wallet-services-sdk/pull/1/files#diff-f69e580bf3fc644bf19e546a0a2ae0cd8488e526fbe4e7b24a6b7a81b8df2a7eR69-R71

won't need the resolve properties bit since the type here isn't something with `Promise` fields. But the deeplyHexlify bit is needed: https://github.com/OMGWINNING/wallet-services-sdk/blob/36af2466dbe2afdfbee8cc8e60acfa348e379064/packages/uo-builder/src/deeplyHexlify.ts

src/api/bundler-namespace.ts
```diff
@@ -0,0 +1,147 @@
...
+      userOperation.callGasLimit = gasEstimates.callGasLimit;
+      userOperation.preVerificationGas = gasEstimates.preVerificationGas;
+      userOperation.verificationGasLimit = gasEstimates.verificationGasLimit;
+    }
+
+    if (!userOperation.maxFeePerGas || !userOperation.maxPriorityFeePerGas) {
+      const feeData = await provider.getFeeData();
+      userOperation.maxFeePerGas = feeData.maxFeePerGas;
+      userOperation.maxPriorityFeePerGas = feeData.maxPriorityFeePerGas;
+    }
```
moldy530, line 51: ah h/u this is my bad, we can't do this because it changes the hash which will impact the signature

PR #332, opened 2023-05-30
Ay/acquired at

ayang-rkjs:


src/api/nft.ts
```diff
@@ -106,4 +107,7 @@ export interface Nft extends BaseNft {
 
   /** Detailed information on why an NFT was classified as spam. */
   spamInfo?: SpamInfo;
+
+  /** Time at which the NFT was most recently acquired by the user. */
+  acquiredAt?: AcquiredAt;
```
ayang-rkjs, line 112: cool to use this type in both raw and user-facing response?

thebrianchen: let's keep them separate. it's in a little messy in v2, but raw and user-facing interfaces are all separate in v3

src/types/types.ts
```diff
@@ -598,6 +598,13 @@ export interface SpamInfo {
   classifications: NftSpamClassification[];
 }
 
+/** Block time and number at which an NFT was acquired. */
+export interface AcquiredAt {
+  blockTimestamp: string;
```
thebrianchen, line 603: can you add comments for each field?

src/internal/raw-interfaces.ts
```diff
@@ -31,12 +31,12 @@ export interface RawSpamInfo {
   classifications: NftSpamClassification[];
 }
 
-// Information on the time at which an NFT was last acquired.
+/** Information on the time at which an NFT was last acquired. */
 export interface RawAcquiredAt {
-  // Timestamp of the block at which an NFT was last acquired.
+  /** Timestamp of the block at which an NFT was last acquired. */
   blockTimestamp: string;
```
thebrianchen, line 37: Missed this the first review but these fields can be undefined in v2 right?

src/types/types.ts
thebrianchen: These fields should be optional I think

thebrianchen:
Can you also add a changelog entry?

src/types/types.ts
ayang-rkjs: u right

src/internal/raw-interfaces.ts
ayang-rkjs: yup

src/types/types.ts
```diff
@@ -600,6 +600,15 @@ export interface SpamInfo {
   classifications: NftSpamClassification[];
 }
 
+/** Block time and number at which an NFT was acquired. */
```
thebrianchen, line 603: Add a comment that this is only available when specifying the orderby = transfertime.

src/api/nft.ts
thebrianchen: Add a comment that this is only available when specifying the orderby = transfertime pls

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Minor Changes
 
 - Added redundancy to `CoreNamespace.getTokensForOwner()` to handle failures when fetching token metadata.
+- Added support for the `acquiredAt` field for ordered `getNftsForOwner` queries.
```
thebrianchen, line 10: Add a comment that this is only available when specifying the orderby = transfertime.

thebrianchen (approved):
lgtm with another request! sorry for the all the one off changes

PR #335, opened 2023-05-31
add validAt field to getNFTSales v2 response

alex-miao:


thebrianchen (approved):
thanks for adding this to the SDK! can you also include a changelog entry (minor) in CHANGELOG.md?

PR #340, opened 2023-06-07
add refreshCache

ayang-rkjs:


src/api/nft-namespace.ts
```diff
@@ -523,9 +523,10 @@ export class NftNamespace {
    */
```
thebrianchen, line 523: add jsdoc `@param` here and in summarizeNftAttributes

thebrianchen (approved):
can you also add a changelog entry that describes what it does // default behavior?

PR #343, opened 2023-06-22
06 22 include count

ayang-rkjs:


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Add the `refreshCache` parameter to NFT rarity endpoints to allow users to update stale rarity values.
```
thebrianchen, line 5: move this into minor changes pls

src/types/types.ts
```diff
@@ -851,6 +851,9 @@ export interface GetOwnersForNftResponse {
 export interface GetOwnersForContractResponse {
   /** An array of owner addresses for the provided contract address */
   owners: string[];
+
+  /** Total count of unique owners */
+  totalCount: number;
```
thebrianchen, line 856: this should be optional right? 

Can you also add a comment here that to get this field you have to set `{@link GetOwnersForContractOptions.includeCount`} to true?

thebrianchen (approved):
thanks for plumbing through

PR #360, opened 2023-08-23
base sdk

kunal365roy:
Base is an OP Stack chain and supports all the same deps as optimism chains. Its supported in ethers as well per their docs, however the tests didn't seem to pass when I didn't blacklist it. Alchemy has support for all core rpc methods at time of writing listed here: https://alchemyinsights.slack.com/archives/CD5H0MAFM/p1691717149204729

w/ some followup timelines for enhanced features seen here: https://www.notion.so/alchemotion/Internal-Only-Base-GTM-Run-of-Show-605f8a5e945f477796d94ad46bd4b884

### DIFF
--- a/src/util/const.ts
+++ b/src/util/const.ts
@@ -88,6 +88,14 @@ export const CustomNetworks: { [key: string]: NetworkFromEthers } = {
   'polygonzkevm-testnet': {
     chainId: 1442,
     name: 'polygonzkevm-testnet'
+  },
+  'base-mainnet': {
+    chainId: 8453,
+    name: 'polygonzkevm-mainnet'
+  },
+  'base-goerli': {
+    chainId: 84531,
+    name: 'polygonzkevm-testnet'
   }
 };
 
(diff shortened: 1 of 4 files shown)


src/util/const.ts
```diff
@@ -56,7 +56,11 @@ export const EthersNetwork = {
   [Network.MATIC_MUMBAI]: 'maticmum',
   [Network.ASTAR_MAINNET]: 'astar-mainnet',
   [Network.POLYGONZKEVM_MAINNET]: 'polygonzkevm-mainnet',
-  [Network.POLYGONZKEVM_TESTNET]: 'polygonzkevm-testnet'
+  [Network.POLYGONZKEVM_TESTNET]: 'polygonzkevm-testnet',
+  [Network.POLYGONZKEVM_MAINNET]: 'polygonzkevm-mainnet',
```
thebrianchen, line 60: pls delete duplicate polygon entries

src/util/const.ts
```diff
@@ -84,6 +88,14 @@ export const CustomNetworks: { [key: string]: NetworkFromEthers } = {
   'polygonzkevm-testnet': {
     chainId: 1442,
     name: 'polygonzkevm-testnet'
+  },
+  'base-mainnet': {
+    chainId: 8453,
+    name: 'polygonzkevm-mainnet'
```
thebrianchen, line 94: update the name fields here and below

thebrianchen (changes requested):

src/util/const.ts
kunal365roy: ah sorry, weird thought this might've failed unit tests

thebrianchen (approved):

PR #394, opened 2023-11-30
feat(opt-base-sepolia): adding support for sepolia networks on opt an…

rambhask:
…d base


thebrianchen (commented):
looks good! can you also leave an entry in changelog.md?

thebrianchen (approved):

PR #402, opened 2023-12-22
adding base networks for notify namespace

StoyanD:
Adding the base networks to the notify namespace map so that webhooks work and the sdk no longer throws when trying to work with the base chains


avarobinson (commented):
LGTM

dphilipson (approved):

StoyanD:
any idea when this is going to make it into a release?

rambhask (approved):

PR #403, opened 2024-01-08
feat: add reportSpam and isAirdropNFT endpoints

avasisht23:
https://alchemyinsights.slack.com/archives/C04P0BZDUQ5/p1704678527884079

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,7 +226,7 @@ under the `alchemy.nft` namespace:
 - `isSpamContract()`: Check whether the given NFT contract address is a spam contract as defined by Alchemy (see the [NFT API FAQ](https://docs.alchemy.com/alchemy/enhanced-apis/nft-api/nft-api-faq#nft-spam-classification))
 - `getSpamContracts()`: Returns a list of all spam contracts marked by Alchemy.
 - `reportSpam()`: Report feedback that a given NFT contract address is a spam contract as defined by Alchemy.
-- `isSpamContract()`: Check whether the given NFT token is marked as an airdrop or not. Airdrops are defined as NFTs that were minted to a user address in a transaction sent by a different address.
+- `isAirdropNft()`: Check whether the given NFT token is marked as an airdrop or not. Airdrops are defined as NFTs that were minted to a user address in a transaction sent by a different address.
 - `refreshNftMetadata()`: Refresh the cached NFT metadata for a contract address and a single tokenId.
 - `refreshContract()`: Enqueues the specified contract address to have all token ids' metadata refreshed.
 - `getFloorPrice()`: Return the floor prices of a NFT contract by marketplace.
--- a/README.md
+++ b/README.md
@@ -225,6 +225,8 @@ under the `alchemy.nft` namespace:
 - `verifyNftOwnership()`: Check whether the provided owner address owns the provided NFT contract addresses.
 - `isSpamContract()`: Check whether the given NFT contract address is a spam contract as defined by Alchemy (see the [NFT API FAQ](https://docs.alchemy.com/alchemy/enhanced-apis/nft-api/nft-api-faq#nft-spam-classification))
 - `getSpamContracts()`: Returns a list of all spam contracts marked by Alchemy.
+- `reportSpam()`: Report feedback that a given NFT contract address is a spam contract as defined by Alchemy.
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -225,6 +225,8 @@ under the `alchemy.nft` namespace:
 - `verifyNftOwnership()`: Check whether the provided owner address owns the provided NFT contract addresses.
 - `isSpamContract()`: Check whether the given NFT contract address is a spam contract as defined by Alchemy (see the [NFT API FAQ](https://docs.alchemy.com/alchemy/enhanced-apis/nft-api/nft-api-faq#nft-spam-classification))
 - `getSpamContracts()`: Returns a list of all spam contracts marked by Alchemy.
+- `reportSpam()`: Report feedback that a given NFT contract address is a spam contract as defined by Alchemy.
+- `isSpamContract()`: Check whether the given NFT token is marked as an airdrop or not. Airdrops are defined as NFTs that were minted to a user address in a transaction sent by a different address.
```
thebrianchen, line 229: rename `isSpamContract` to `isAirdropNft`

src/api/nft-namespace.ts
```diff
@@ -473,6 +477,33 @@ export class NftNamespace {
...
+
+  /**
+   * Returns whether a token is marked as an airdrop or not.
+   * Airdrops are defined as NFTs that were minted to a user address in a transaction
+   * sent by a different address.
+   *
+   * @param contractAddress - The contract address to check.
+   * @beta
+   */
+  isAirdropNFT(
```
thebrianchen, line 500: rename `NFT` to `Nft` to stay consistent with style

src/api/nft-namespace.ts
```diff
@@ -473,6 +477,33 @@ export class NftNamespace {
     return getSpamContracts(this.config);
   }
 
+  /**
+   * Returns whether a contract is marked as spam or not by Alchemy. For more
+   * information on how we classify spam, go to our NFT API FAQ at
+   * https://docs.alchemy.com/alchemy/enhanced-apis/nft-api/nft-api-faq#nft-spam-classification.
+   *
+   * @param contractAddress - The contract address to check.
+   * @beta
```
thebrianchen, line 486: we don't need the `@beta` tags here since they are already released into the NFT API endpoints.

src/api/nft-namespace.ts
```diff
@@ -473,6 +477,33 @@ export class NftNamespace {
...
+   */
+  reportSpam(contractAddress: string): Promise<ReportSpamResponse> {
```
thebrianchen, line 488: Since the return string of a successful response is always just a message indicating the NFT was reported successfully, can we return `Promise<void>` in this method and just not return the underlying string?

We can also delete `ReportSpamResponse`.

thebrianchen (changes requested):
Thanks for taking this on and adding tests too!

Can you also add a CHANGELOG.md entry as well? 🙏

src/internal/nft-api.ts
```diff
@@ -622,6 +624,40 @@ export async function getSpamContracts(
...
+      contractAddress
+    }
+  );
+}
+
+export async function isAirdropNft(
+  config: AlchemyConfig,
+  contractAddress: string,
+  tokenId: string,
+  srcMethod = 'isAirdropNFT'
```
thebrianchen, line 647: nit: rename to `Nft`

src/internal/raw-interfaces.ts
```diff
@@ -385,6 +385,10 @@ export interface RawGetSpamContractsResponse {
   contractAddresses: string[];
 }
 
+export interface RawisAirdropNftResponse {
```
thebrianchen, line 388: nit `RawIs` instead of `Rawis`

src/types/nft-types.ts
```diff
@@ -384,6 +384,10 @@ export interface IsSpamContractResponse {
   isSpamContract: boolean;
 }
 
+export interface isAirdropNftResponse {
```
thebrianchen, line 387: nit: capitalize to `Is`

thebrianchen (approved):
some final nits, thanks!

src/internal/nft-api.ts
avasisht23: doesn't this one need to be the actual endpoint? which is spelled [isAirdropNFT](https://docs.alchemy.com/reference/isairdropnft-v3)?

avasisht23: oh i see this is the one that gets logged, not the restApiName

Issue #416, opened 2024-04-02
 Cannot read properties of undefined (reading 'safelistRequestStatus')


JohnPochta:
### [REQUIRED] Environment

- Alchemy SDK version: 3.2.0

### [REQUIRED] Describe the problem
I have the code snippet:

```
this.alchemy = new Alchemy({
    url: this.NETWORK_RPC_URL,
  });

return await this.alchemy.nft.getNftsForOwner(address, {
  omitMetadata: false,
});
```

This throws the error:  Cannot read properties of undefined (reading 'safelistRequestStatus') .
I figured out that it's related to OpenSea metadata parsing and most likely someone forgot smth like: 
`
rawNftContract.openSeaMetadata**?**.safelistRequestStatus 
`
in src/utils/util.ts.

#### How to reproduce:

```
this.alchemy = new Alchemy({
      url: this.NETWORK_RPC_URL,
    });

return await this.alchemy.nft.getNftsForOwner(address, {
  omitMetadata: false,
});
```

nirmit1509:
@JohnPochta Can you please escalate this issue. There is an open PR for the same as well. Thanks in advance 🙏

thebrianchen:
Sorry for the delay. This fix has been applied and published in `v3.3.1`.

nirmit1509:
https://github.com/alchemyplatform/alchemy-sdk-js/pull/432

nirmit1509:
@thebrianchen The comparison with null was also wrong.
Can you please fix this as well.
https://github.com/alchemyplatform/alchemy-sdk-js/pull/432

anthhub:
@thebrianchen The comparison with null was also wrong.
Can you please fix this as well.
https://github.com/alchemyplatform/alchemy-sdk-js/pull/432

thebrianchen:
@anthhub @nirmit1509 Sorry for the delay in not fixing this. This is now fixed in `v3.4.1`.

PR #419, opened 2024-04-15
fix: SafelistRequestStatus bug fix issue #416

nirmit1509:
This PR fixes the `Cannot read properties of undefined (reading 'safelistRequestStatus')` issue.
Added a conditional operator to fix this.


thebrianchen:
Fixes #416.

PR #438, opened 2024-07-31
Create  get the latest block const latestBlock = alchemy.core.getBloc…curl https://eth-mainnet.g.alchemy.com/v2/BfJQwDGregJAzM6y31zXE_lN8c52sfmo -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["latest", false],"id":0}'

Wadalzalat:
…k("latest").then(console.log);


Wadalzalat:
package main

import (
"context"
"fmt"
"log"

"github.com/ethereum/go-ethereum/common"
"github.com/ethereum/go-ethereum/ethclient"
)

func main() {
  client, err := ethclient.Dial("https://eth-mainnet.g.alchemy.com/v2/BfJQwDGregJAzM6y31zXE_lN8c52sfmo")
  if err != nil {
      log.Fatal(err)
  }

  // Get the latest known block
  block, err := client.BlockByNumber(context.Background(), nil)
  if err != nil {
      log.Fatal(err)
  }
  fmt.Println("Latest block:", block.Number().Uint64())
}

PR #454, opened 2024-09-27
add overrides for elliptic and ws packages to address security concerns

SahilAujla:
# Overrides Discussion
Initially when I added the overrides, I was trying to update the `yarn.lock` file but then I realized we're using yarn v1 which does not support overrides. But if we update to yarn v2+ that will introduce significant changes to the project like: 

- replacing `node_modules` with Plug n Play (PnP)
- introducing new files such as `.pnp.cjs`
- introduction of `.yarn` directory etc.

And it would completely change the `yarn.lock` file. So I went with [`resolutions`](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/) to achieve the same result that that also works with yarn v1.

# Changes

- Addressed security vulnerabilities by adding overrides (resolutions) for `elliptic` and `ws` dependencies. Specifically:
  - Updated `elliptic` to version `6.5.7` (latest) to fix vulnerabilities related to EDDSA signature length checks and ECDSA BER-encoded signatures.
  - Updated `ws` to version `7.5.10` to mitigate a potential DoS vulnerability. The latest major version of `ws` is `8.x`, however, `7.5.10` is the most up-to-date version compatible with `ethers v5`, which depends on `7.x` versions of `ws`.


thebrianchen (approved):
looks good, thanks!

PR #463, opened 2024-10-21
fix: modify webhook network map to include all supported networks

vahnag:


src/api/notify-namespace.ts
```diff
@@ -635,25 +635,8 @@ export class NotifyNamespace {
...
-  OPT_GOERLI: Network.OPT_GOERLI,
-  OPT_SEPOLIA: Network.OPT_SEPOLIA,
-  BASE_MAINNET: Network.BASE_MAINNET,
-  BASE_GOERLI: Network.BASE_GOERLI,
-  BASE_SEPOLIA: Network.BASE_SEPOLIA,
-  ZKSYNC_MAINNET: Network.ZKSYNC_MAINNET,
-  ZKSYNC_SEPOLIA: Network.ZKSYNC_SEPOLIA
-};
+const WEBHOOK_NETWORK_TO_NETWORK: { [key: string]: Network } =
+  Object.fromEntries(Object.entries(Network));
```
stanleyjones, line 639: Why does this work? Were we not including all the networks in this list?

vahnag: Yes, it generates the same map from the Network enum on the fly. 
This will ensure that we do not forget to add networks here in the future.

We could probably refactor this to remove these maps altogether and use the Network enum directly, but wanted to keep the changes to the minimum.

Screenshot 2024-10-22 at 00 56 59

vahnag: @thebrianchen the only thing that this object does to prevent that is replace the network field in the request with `undefined` before making webhook requests. 

It's not like it's doing any validation before making the requests.

Why not just pass on the unsupported network to the webhooks service and let it handle the error instead of nullifying the network field?

thebrianchen: I think they wanted to release gate some of the webhooks, and what i had here was definitely not the best way to do it.

thebrianchen (approved):
thanks for cleaning this up!

PR #488, opened 2025-03-24
feat: Add Monad, Settlus.

stanleyjones:
See title.


CHANGELOG.md
```diff
@@ -6,6 +6,15 @@
 
 ### Minor Changes
 
+## 3.5.6
+
+### Minor Changes
+
+- Added `MONAD` testnet.
+- Added `SETTLUS` testnet (Sepolia).
+- Deprecated `BERACHAIN` testnet (Bartio).
```
bmoyroud, line 15: Do we want to deprecate this chain?

CHANGELOG.md
```diff
@@ -6,6 +6,15 @@
...
+- Deprecated `CELO` mainnet, testnet (Alfajores).
```
bmoyroud, line 16: Do we want to deprecate Celo? They are moving to an L2 tomorrow & we'll be supporting.

src/types/types.ts
```diff
@@ -145,11 +145,14 @@ export enum Network {
   BNB_TESTNET = 'bnb-testnet',
   AVAX_MAINNET = 'avax-mainnet',
   AVAX_FUJI = 'avax-fuji',
+  /** @deprecated */
```
bmoyroud, line 148: We'll be supporting this chain - would not mark as deprecated.

bmoyroud (commented):
We need to keep Berachain and Celo in SDK.

CHANGELOG.md
stanleyjones: I believe mainnet is still supported.

CHANGELOG.md
stanleyjones: Will they have the same chain ID, etc. as an L2?

stanleyjones:
Removed deprecations to not block Monad.

SahilAujla (approved):

PR #492, opened 2025-04-10
add webhook naming via sdk

GregoryElliott:
add name as an option to the create-webhook api SDK - the name parameter already exists on the create-webhook alchemy api

ASANA TASK: https://app.asana.com/1/1129441638109975/project/1208969177908953/task/1209648386732973?focus=true

### DIFF
--- a/src/api/notify-namespace.ts
+++ b/src/api/notify-namespace.ts
@@ -55,7 +55,7 @@ import { AlchemyConfig } from './alchemy-config';
  *
  * Do not call this constructor directly. Instead, instantiate an Alchemy object
  * with `const alchemy = new Alchemy(config)` and then access the notify
- * namespace via `alchemy.notify`.
+ * namespace via `alchemy. `.
  */
 export class NotifyNamespace {
   /** @internal */
@@ -519,6 +519,7 @@ export class NotifyNamespace {
       webhook_type: type,
       webhook_url: url,
       ...(appId && { app_id: appId }),
+      ...(params.name && { name: params.name }),
 
       // Only include the filters/addresses in the final response if they're defined
       ...nftFilterObj,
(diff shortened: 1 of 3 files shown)


src/api/notify-namespace.ts
```diff
@@ -55,7 +55,7 @@ import { AlchemyConfig } from './alchemy-config';
  *
  * Do not call this constructor directly. Instead, instantiate an Alchemy object
  * with `const alchemy = new Alchemy(config)` and then access the notify
- * namespace via `alchemy.notify`.
+ * namespace via `alchemy. `.
```
vahnag, line 58: Why change this?

src/api/notify-namespace.ts
```diff
@@ -667,7 +668,9 @@ function parseRawWebhook(rawWebhook: RawWebhook): Webhook {
     signingKey: rawWebhook.signing_key,
     version: rawWebhook.version as WebhookVersion,
     // Only include the appId in the final response if it's defined
-    ...(rawWebhook.app_id !== undefined && { appId: rawWebhook.app_id })
+    ...(rawWebhook.app_id !== undefined && { appId: rawWebhook.app_id }),
+    // Only include the name in the final response if it's defined
+    ...(rawWebhook.name !== undefined && { name: rawWebhook.name })
```
vahnag, line 673: You might not need this extra check as all webhook types can have a name.

src/types/types.ts
```diff
@@ -1310,6 +1318,8 @@ export interface AddressWebhookParams {
    * created on network of the app provided in the api key config.
    */
   network?: Network;
+  /** Optional name for the webhook. */
+  name?: string;
```
vahnag, line 1322: How do you feel about defining a base CommonWebhookParams with name property that all of these interfaces would extend?

src/api/notify-namespace.ts
GregoryElliott: whoops nice catch

src/types/types.ts
GregoryElliott: i like that - updated

PR #500, opened 2025-04-21
Publishing next version

GregoryElliott:
- Added Solana address activity webhook
- Added animation to NFT type
- Added name to create-webhook


src/api/notify-namespace.ts
```diff
@@ -2,7 +2,6 @@ import { AxiosRequestConfig, Method } from 'axios';
 
 import { BigNumber } from '@ethersproject/bignumber';
 
-import { isAddress } from '@solana/web3.js';
```
GregoryElliott, line 5: making this change required upgrading the solana/web3 package, which created cascading errors so I decided to undo this to unblock the other changes in this published version

src/types/types.ts
```diff
@@ -16,6 +16,9 @@ import {
 
 export * from './ethers-types';
 
+/** @internal Workaround to prevent TS unused import error */
```
GregoryElliott, line 19: typescript was complaining without this

ddgromit (approved):

PR #506, opened 2025-05-14
Add missing originalUrl from NFT animation type

0xmichalis:
This looks like an omission since the API returns `originalUrl` as part of animations
Screenshot from 2025-05-14 17-45-24


thebrianchen (approved):
thanks for adding this!

PR #517, opened 2025-05-23
fix: Don't force batch operations for `getTokensForOwner`

stanleyjones:
# Overview

Not all chains support batch operations so `getTokensForOwner` was failing on some of the newer ones. This PR removes the `forceBatch` setting.


vahnag (approved):

stanleyjones:
Okay, confirmed this is working now.

PR #535, opened 2025-09-12
Fix reportSpam to pass isSpam and return typed response

0xmichalis:


stanleyjones:
Please reopen without the docs changes.

0xmichalis:
@stanleyjones done

